PR #5, opened 2021-01-29
feat: add fauna adapter

pbteja1998:


packages/fauna/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "Bhanu Teja P",
+  "license": "ISC",
+  "peerDependencies": {
+    "next-auth": "^3.1.0",
+    "faunadb": "^4.0.2"
+  },
+  "dependencies": {
+    "crypto": "^1.0.1"
```
balazsorban44, line 16: From npm:
> This package is no longer supported. It's now a built-in Node module. If you've depended on crypto, you should switch to the one that's built-in.

In other words, I think it is unnecessary to add it as a dependency.

packages/fauna/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "@next-auth/fauna-adapter",
+  "version": "0.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
```
balazsorban44, line 8: Could you add these to package.json:

```json
"private": false,
"publishConfig": {
  "access": "public"
},
```
It seems t be required to be able to publish scoped packages.

packages/fauna/index.js
```diff
@@ -0,0 +1,537 @@
+import { query as q } from 'faunadb'
```
balazsorban44, line 1: I think this file should be transpiled to something(?), since it is now an individual package consumed by other projects. @williamluke4 do you have a suggestion? Could we use `tsc` as a `build` script to do this, since it is used already for the Prisma Adapter? In that case, does it need to be installed in `devDependencies` for all packages, or can we install it for all at the same time with `lerna` somehow?

balazsorban44 (changes requested):
Thank you so much @pbteja1998 for this PR!

As you may have seen, we are still figuring out some things in the workflow, so there might be some small changes required before this is merged, but I think it is already good work. 👏

matha-io:
Hello ! Can I help in any way ?

PR #17, opened 2021-01-31
feat: add original typeorm adapter 

ndom91:
Added original typeorm adapter, including adjusting tests, etc.


packages/typeorm/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "mongodb": "^3.6.3",
+    "mssql": "^6.3.1",
+    "mysql": "^2.18.1",
+    "next-auth": "^3.1.0",
+    "pg": "^8.5.1",
+    "standard": "^16.0.3",
+    "typeorm-adapter": "^1.2.0"
+  },
+  "peerOptionalDependencies": {
+    "next-auth": "^3.1.0",
```
balazsorban44, line 52: I think `next-auth` should be under `peerDependencies`

balazsorban44 (changes requested):
So I went through the code, and I assumed that most of it are just a copy+paste from the original repo, so I did not check the code very carefully.

Could you maybe add a README.md to the package with some information about how to use it? Otherwise, I think it is ok.

After `next-auth@3.2.0` is out, I think we can drop the code from the original repo in the next `canary`, and replace it with this package by adding it to `dependencies`.

packages/typeorm/package.json
ndom91: Yeah exactly, I didnt change any of the internal typeorm stuff. That was all taken 1 to 1 from the original adapter code. 

I'll add a readme and update the peer dependency stuff now and then yeah once we have this repo organised and looking good we can focus on taking the adapters out of the primary repo :+1:

packages/typeorm/README.md
```diff
@@ -0,0 +1,92 @@
...
+This is the TypeORM Adapter for [`next-auth`](https://next-auth.js.org). This package can only be used in conjunction with the primary `next-auth` package. It is not a standalone package.
+
+## Getting Started
+
+1. Install `next-auth` and `@next-auth/typeorm-adapter`
+```js
+npm install --save-prod next-auth @next-auth/typeorm-adapter
+```
+
+2. Add this adapter to your `/src/pages/api/[...nextauth].js` next-auth configuration object.
```
balazsorban44, line 30:
```suggestion
2. Add this adapter to your `pages/api/[...nextauth].js` next-auth configuration object.
```

I think it is clear enough only writing the path beginning with `pages`. I cannot speak for everyone else, but I for example don't put my pages under `src`. I am pretty sure that most of the Next.js users are also aware that it can be either src or the root folder.

packages/typeorm/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Overview
+
+This is the TypeORM Adapter for [`next-auth`](https://next-auth.js.org). This package can only be used in conjunction with the primary `next-auth` package. It is not a standalone package.
+
+## Getting Started
+
+1. Install `next-auth` and `@next-auth/typeorm-adapter`
+```js
+npm install --save-prod next-auth @next-auth/typeorm-adapter
```
balazsorban44, line 27: Didn't actually know about the `--save-prod` flag. Looks like it is because it is the default behavior https://docs.npmjs.com/cli/v6/commands/npm-install and I have never had to use it. Could we remove it?

packages/typeorm/README.md
```diff
@@ -0,0 +1,92 @@
...
+   </p>
+   <p align="center" style="align: center;">
+      <img src="https://github.com/nextauthjs/adapters/workflows/Build%20Test/badge.svg" alt="Build Test" />
+      <img src="https://github.com/nextauthjs/adapters/workflows/Integration%20Test/badge.svg" alt="Integration Test" />
+      <img src="https://img.shields.io/bundlephobia/minzip/@next-auth/typeorm-adapter" alt="Bundle Size"/>
+      <img src="https://img.shields.io/npm/dm/@next-auth/typeorm-adapter" alt="Downloads" />
+      <img src="https://img.shields.io/github/stars/nextauthjs/adapters" alt="Github Stars" />
+      <img src="https://img.shields.io/github/v/release/nextauthjs/adapters?include_prereleases" alt="Github Release" />
+   </p>
+</p>
```
balazsorban44, line 17: Should we enforce a header like this on all the adapters?

Should we enforce certain sections in the README files in general?

Nonetheless we should make sure that the example adapter contains as much boilerplate for an easy start, as possible! https://github.com/nextauthjs/adapters/tree/canary/packages/example is a bit empty right now.

balazsorban44 (changes requested):

packages/typeorm/README.md
ndom91: Yeah its the same as `-s`, I just usually like to be explicit about these types of things in readmes. But seeing as its the default behavior anyway yeah lets just remove it.

packages/typeorm/README.md
ndom91: Hmm yeah the headers like are probably not necessary on all the adapters. Ill see if I can whip up a simplified version.

What exactly is the goal of the example adapter actually?

ndom91: I would like to keep some basic badges for the adapters though, like 1) tests passing/failing 2) minzipped size of adapter pkg and 3) latest release version. What do you think?

balazsorban44: I think the goal of the example adapter should be to be able to simply copy+paste it and get your adapter working as fast as possible. All the config, known dependencies, readme, everything is filled out for you.

balazsorban44: We should discuss the example adapter internally first. You corrected my concerns on the `typeorm-adapter` for now.

balazsorban44 (approved):
LGTM!

PR #19, opened 2021-01-31
feat: add faunadb adapter

ndom91:
Added the previously removed faunadb adapter. This is based off the status of the adapter at the time of the commit where you removed it from the main repo, so it should be the latest version of this code. I haven't tested the test, docker container, etc. this is just to get the basics into this repo and get a starting point to work with :+1: 

CC: @Joev-, @waptik, @n44ps, @perkinsjr - we're adding this to its own repo/package now if any of you have any interest in contributing :)


ndom91:
Ah sorry didn't see that there was already a PR open for the old fauna adapter!

PR #18, opened 2021-01-31
feat(firebase): add firebase adapter

ndom91:
Added the initial firebase adapter folder structure / content based off of @RonHouben's adapter (https://github.com/RonHouben/nextjs-tailwindcss-todo-app/blob/c55af99329f5206d31eec8f87405873e111895c6/lib/firebaseAdapter.ts). 

Status: ts builds successfully, jest is setup just like in the prisma adapter, however there are obviously no tests written yet.

Closes #84

PR #36, opened 2021-02-17
add: tests for firestore adapter

AlexMeuer:
Ported the Prisma tests.

Relates to #18

### DIFF
--- a/packages/firebase/tests/index.test.ts
+++ b/packages/firebase/tests/index.test.ts
@@ -1,1 +1,153 @@
-/* Placeholder for Firebase Tests */
+import * as admin from "firebase-admin";
+import { exposeMockFirebaseAdminApp } from 'ts-mock-firebase';
+import Adapter, {
+    IUser,
+    ISession,
+    IVerificationRequest
+} from "../src";
+
+const app = admin.initializeApp({});
+const mocked = exposeMockFirebaseAdminApp(app);
+const firebaseAdapter = Adapter({
+    firestoreAdmin: admin.firestore(),
+    usersCollection: "users",
+    accountsCollection: "accounts",
+    sessionsCollection: "sessions",
+    verificationRequestsCollection: "verificationRequests"
+});
+
+let session: ISession | null = null;
+let user: IUser | null = null;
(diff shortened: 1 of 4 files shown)


packages/firebase/tests/index.test.ts
```diff
@@ -1 +1,153 @@
...
+            {
+                ...session,
+                expires: expiresRequested,
+            },
+            true
+        )) as ISession;
+
+        const difference = Math.abs(session.expires.getTime() - expiresExpected);
+        expect(difference).toBeLessThan(5);
+    });
```
AlexMeuer, line 95: Unlike Prisma, the Firestore adapter does not return the session with the given expiry time. This is the test I'm least sure about.

ndom91:
@AlexMeuer thanks a lot for getting this done! One problem though, did you have `ts-mock-firebase` installed locally? I wasn't included in the package.json, but is required by some of your tests.

I tried to install myself, but unfortunately it won't compile on my linux install, its trying to compile from source this `grpc` library from python I think, so theres all sorts of node-gyp errors. There hasn't been a release in 2 years for `ts-mock-firebase` unfortunately so I assume some versions its requiring are too old / incompatible with current python / node-gyp / whatever

AlexMeuer:
Hmm, I must have forgotten to save the dependency. I'll have a look for an alternative when I've time.

AlexMeuer:
Spent some time trying to replace `ts-mock-firebase` with `firebase-mock` but it lacks support for serverTimestamp as well as screwing with async/await. Might just have to manually mock the used methods, but it doesn't look like I'll have more time this week.

ndom91:
Hey so first of all, thanks for giving it a shot and reporting back on any and all stumbling blocks! 

I'll take a shot at specifically mocking the used methods then. Dom have you pushed your latest state to this branch?

balazsorban44:
Thank you for this, but I think https://github.com/nextauthjs/adapters/pull/90 will address most of this, and also some of the stuff here has been made redundant by improving our core package. It takes care of error handling instead of the adapters.

PR #45, opened 2021-04-11
fix: Add identifier to verification requests

WillTheVideoMan:
CC: @balazsorban44 

This PR includes the `identifier` in the Fauna query for `getVerificationRequest` and `deleteVerificationRequest`. This prevents a valid token being used in combination with any identifier other than the one for which the request was issued.

Also change the documentation to reflect the necessary change to the `verification_request_by_token` index to include the identifier. 

This should be treated as a breaking change since an update to a Fauna index is required, which required manual intervention.

### DIFF
--- a/packages/fauna/README.md
+++ b/packages/fauna/README.md
@@ -32,6 +32,9 @@ CreateIndex({
   name: 'verification_request_by_token',
   source: Collection('verification_requests'),
   unique: true,
-  terms: [{ field: ['data', 'token'] }],
+  terms: [
+    { field: ['data', 'token'] }, 
+    { field: ['data', 'identifier'] }
+  ],
 })
 ```
--- a/packages/fauna/src/index.js
+++ b/packages/fauna/src/index.js
@@ -448,7 +448,10 @@ const Adapter = (config, options = {}) => {
         .digest('hex')
       const FQL = q.Let(
         {
-          ref: q.Match(q.Index(indexes.VerificationRequest), hashedToken),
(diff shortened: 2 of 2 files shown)


pbteja1998 (approved):

pbteja1998:
@WillTheVideoMan I think it would be good to setup Fauna migrations(may be in a separate PR).

I use [fauna-schema-migrate](https://github.com/fauna-brecht/fauna-schema-migrate) package and it works well for me so far. It's created by a developer advocate at Fauna, but it's not yet official.

What do you think?

pbteja1998:
@WillTheVideoMan Thanks a lot for bringing this to notice and fixing it. Appreciate it very much.

pbteja1998:
Shouldn't we also change the version number of the adapter? Or does that happen automatically with the monorepo setup?

balazsorban44:
Yeah.. The monorepo setup is a bit clunky at the moment, but it SHOULD bump versions automatically. We will have to review if we want to continue with lerna, or we should try to find a (hopefully) easier alternative. Unfortunately I don't have enough time for the adapters at the moment. 😕

balazsorban44 (approved):
Thanks so much!

WillTheVideoMan:
@pbteja1998 Hi! Glad I could be of help here. The `fauna-schema-migrate` package looks great, although I wonder if such a tool is right to be included at the package level. 

Maybe instead the tool could be mentioned and recommended in the documentation for this adapter? That way, people can opt-in to its use, and can subsequently have the option to use the tool throughout their full project for migrations further than authentication. 

What do you think?

Issue #38, opened 2021-02-22
uuid or cuid for Prisma adaptor model IDs


jasonkuhrt:
I would like to use `cuid` for model IDs in our Prisma schema.

This will require changing the type of the field.

This in turn forces some of the TS types to change, e.g. `linkAccount` function `userId` paramter changes from `number` to `string`.

Would it be acceptable for force Next Auth Prisma adaptor users to use `String` type for model IDs?

Or is there an easy way to support both `Int` based IDs and `String` based model IDs?

jasonkuhrt:
It seems all that is needed is to do this in the implementation to keep it type safe:

```ts
type UserId = Prisma.User['id']

// replace hard-coded `number` with references to `UserId`
```

kripod:
I just noticed in the [Prisma adapter’s code](https://github.com/nextauthjs/adapters/blob/canary/packages/prisma/src/index.ts) that `number` and `string` aren’t used consistently for `userId`. It seems there shouldn’t be any difference in runtime, but I think those types should be changed as written above.

PR #46, opened 2021-04-13
fix: allow user ID to be passed as a string

sir-dunxalot:
Fixes/supersedes nextauthjs/next-auth#1644

Per [the JWT spec (sub section)](https://tools.ietf.org/html/rfc7519#section-4.1.2), a token's `sub` is a string. Thus, if next-auth is calling `getUser(sub)` ([example here](https://github.com/nextauthjs/next-auth/blob/main/src/server/lib/callback-handler.js#L63)), you're passing a string to getUser(). This PR addresses this incompatibility by allowing `userId` to be passed into the Prisma adapter's methods as either a string or a number.

### DIFF
--- a/packages/prisma/src/index.ts
+++ b/packages/prisma/src/index.ts
@@ -103,22 +103,23 @@ export default function PrismaAdapter<
       }
     }
 
-    async function getUser(id: number) {
-      debug("GET_USER", id);
+    async function getUser(userId: number | string) {
+      const _userId = Number(userId);
+      debug("GET_USER", _userId);
       try {
-        const cachedUser = userCache.get(id);
+        const cachedUser = userCache.get(_userId);
         if (cachedUser) {
           debug("GET_USER - Fetched from LRU Cache", cachedUser);
           // stale while revalidate
           (async () => {
             const user = (await prisma[User as "user"].findUnique({
-              where: { id },
+              where: { id: _userId },
             })) as Prisma.User;
             userCache.set(user.id, user);
           })();
           return cachedUser;
         }
-        return prisma[User as "user"].findUnique({ where: { id } });
+        return prisma[User as "user"].findUnique({ where: { id: _userId } });
       } catch (error) {
         logger.error("GET_USER_BY_ID_ERROR", error);
         // @ts-ignore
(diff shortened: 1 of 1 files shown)


kripod:
Casting each `userId` parameter to `Number` instead of the other way around – `String(userId)` – doesn’t look good to me. IDs are often stored as strings, rather than numbers.

sir-dunxalot:
@kripod I'm not sure how casting to a string would be compatible with the Prisma adapter and schema. The point of this PR is to enable strings and numbers to be passed as the `userId` and turn that value into a number for the Prisma adapter.

Here is the `schema.prisma` file, which shows user ID is stored as int:

image

Maybe you can clarify your comment?

Personally, I would like to use ccid or uuid, which would be strings, as the user ID but #38 shows that isn't possible today.

kripod:
There are no runtime changes to be made. The current code already handles strings and numbers correctly – only the types aren’t correct.

sir-dunxalot:
I think I understand but I would like to make sure we're on the same page... Without these code changes, I encounter a silent error. Specifically, when next-auth gets `sub` from a JWT and calls `getUser()` [like here](https://github.com/nextauthjs/next-auth/blob/main/src/server/lib/callback-handler.js#L63), an error is thrown. However, because of [this empty catch block](https://github.com/nextauthjs/next-auth/blob/main/src/server/lib/callback-handler.js#L66) the error is completely silent. As a result, account linking does not work and a user adding an account to their existing user will end up with a new user created for the new account.

Are you saying we should change, for example, `async function getUser(userId: number)` to `async function getUser(userId: number | string)`. Then cast `userId` to string and the above bug won't be encountered? If so, do we need to cast the value at all? THe only reason I am casting values is because of [this feedback in the original PR](https://github.com/nextauthjs/next-auth/pull/1644#issuecomment-815307297).

If what I just wrote matches your feedback then I will update the PR accordingly.

Thanks!

kripod:
Thanks for your feedback!

I wasn’t aware of the JWT case as I’ve been using a stored session approach all along. In my case, transforming only the types has worked flawlessly. As strings cannot be casted into IEEE 754 numbers without data loss, I would advise against changes like below:

```ts
      const _userId = Number(userId);
```

Casting the other way (from numbers to strings) could work, though, but I’m not sure what the benefit of that would be:

```ts
      const _userId = String(userId);
```

sir-dunxalot:
@kripod Based on [this comment](https://github.com/nextauthjs/adapters/pull/72#issuecomment-828354244) do you still want me to update this PR for merging? If so, I will be happy to make the changes requested above.

kripod:
@sir-dunxalot Since this PR was opened, #72 has landed, which suggests the use of string IDs instead of numbers. I think this is resolved now.

PR #54, opened 2021-04-22
fix(prisma): populating `sessionCache` in `createSession`

kripod:
According to the current implementation of NextAuth, `session` objects must have a `userId` property instead of a `user` with an `id`:

https://github.com/nextauthjs/next-auth/blob/34f334a71d9693ed177168dbc4bf706b2734f8ff/src/server/routes/session.js#L68

The [`/session` route](https://github.com/nextauthjs/next-auth/blob/main/src/server/routes/session.js) of NextAuth only takes the following properties into account on a session object:

- `userId`
- `accessToken`
- `expires`

Also, users are retrieved from the `userCache` when the following line is executed by NextAuth – so this change shouldn’t affect caching performance:

```ts
const user = await getUser(session.userId)
```

### DIFF
--- a/packages/prisma/src/index.ts
+++ b/packages/prisma/src/index.ts
@@ -270,22 +270,14 @@ export default function PrismaAdapter<
         expires = dateExpires.toISOString();
 
         const session = {
+          userId: user.id,
           expires,
           sessionToken: randomBytes(32).toString("hex"),
           accessToken: randomBytes(32).toString("hex"),
-          userId: user.id,
         };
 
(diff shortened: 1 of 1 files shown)


kripod:
This change also makes `sessionToken` and `accessToken` consistent between the DB and the LRU cache. Otherwise there would be token mismatches when a user gets removed from the `sessionCache`.

kripod:
Superseded by https://github.com/nextauthjs/adapters/commit/43221837cf0b21f5a2706b73d2b8dd3ffda863ad

PR #55, opened 2021-04-22
refactor(prisma): loosen `peerDependencies` ranges

kripod:
This change is to ensure better compliance with npm v7’s new peer dependency resolution algorithm.


kripod:
This has been resolved.

PR #62, opened 2021-04-25
chore: reorg repo package.jsons

ndom91:
Reorganised the `package.json`'s (toplevel + package-level) to match that of the main project (`next-auth`). 

This PR also has moved from standard -> eslint/prettier (with standard config) for everything. There is a toplevel config setting in the top `package.json`. The underlying packages will use this configuration as well.

**Edit**: Sorry, not sure the first two commits, `de39e16` and `f20c1b6` snuck into this PR. I tried to cherry pick just the last one into its own branch, but again they came with it. Oh well, its mostly just adding a few base files to form the initial firebase adapter, not changing any existing code.


balazsorban44 (approved):
Awesome! :tada:

PR #72, opened 2021-04-28
feat(prisma): use new Adapter interface

### LINKED ISSUE #38: uuid or cuid for Prisma adaptor model IDs

### LINKED ISSUE #70: [prisma] remove unreliable `sessionCache` and `userCache`, as they don’t work in serverless environments
When lambda function concurrency increases, both `sessionCache` and userCache gets out of sync with the database.

Imagine there are 2 concurrent functions, Alice and Bob, running at the same time:

1. Alice creates a session
2. Bob retrieves the session Alice has just created from the database and caches it in its memory
3. The user logs out and thus, Alice revokes the session by removing it from the database and its own cache
4. Bob still has the given user’s session in its cache, and there’s no way to invalidate it without accessing a synchronization node like a database 😕

I would like to suggest removing the Prisma adapter’s cache in its entirely for now, as the resulting issues tend to cause request timeouts on serverless deployment platforms like Vercel.

balazsorban44:
Currently working out the `id: number` thing...

(#46)

Fixes #70, fixes #38


kripod:
I’ve created a new adapter which works well with string IDs and can be used together with next-auth@3.18.0 out of the box: https://gist.github.com/kripod/35ccf5319c8477de6f71c1877c7de320

packages/prisma/prisma/schema.prisma
```diff
@@ -7,7 +7,7 @@ generator client {
 }
 model Account {
   id                 Int       @id @default(autoincrement())
```
kripod, line 9: I think this should also be converted into a string after all.

packages/prisma/prisma/schema.prisma
```diff
@@ -22,7 +22,7 @@ model Account {
 
 model Session {
   id           Int      @id @default(autoincrement())
```
kripod, line 24: I think all the IDs should be strings for the sake of uniformness.

kripod (approved):
Looks good to me 😊

I would like to suggest using string IDs for `Account` and `Session` inside `schema.prisma`.

packages/prisma/prisma/schema.prisma
balazsorban44: Done

packages/prisma/prisma/schema.prisma
balazsorban44: Done

kripod (approved):

williamluke4 (approved):

PR #90, opened 2021-05-15
test: add adapter agnostic tests [WIP]

balazsorban44:
This PR introduces several tests that are aimed to be agnostic of the underlying adapter. This way, we can ship adapters with higher confidence, and make sure that they all have the same base functionality.

A new `runBasicTests` function is introduced, which can be run at the top of an adapter test file with some very basic configuration, and then a bunch of tests will run, which hopefully tests the functionality that is supposed to be common for all adapters.

Any additional tests can be added after running `runBasicTests`

areas covered in basic tests:
- Meta
  - [x] displayName
- User
  - [x] createUser
  - [x] getUser
  - [x] getUserByEmail
  - [x] updatedUser
  - [ ] ~deleteUser (not implemented in core)~
  - [ ] getUserByProviderAccountId
- Session
  - [x] createSession
  - [x] getSession
  - [x] updateSession
  - [x] deleteSession
- Account
  - [ ] ~unlinkAccount (not implemented in core)~
  - [ ] linkAccount
- Verification Request
  - [x] createVerificationRequest
  - [x] getVerificationRequest
  - [x] deleteVerificationRequest

For now, I am going to add these tests to the currently existing adapters:

- [x] typeorm-legacy 
- [x] prisma-legacy
- [x] prisma
- [x] fauna
~- [ ] dynamodb~


packages/fauna/tests/resources/collections/users.fql
```diff
@@ -0,0 +1,41 @@
...
+#   otherDetails: {
+#     "userType": String('developer' | 'employer' | 'none')
+#     "mobile": String
+#     "isCurrentlyWorking": String('yes' | 'no' | 'none')
+#     "lookingForWork": String('yes' | 'no' | 'none')
+#     "company": String
+#     "technologiesFamiliarWith": String[]
+#   }
+# }
+**/
```
pbteja1998, line 41: I mistakenly copied these from other files. We don't actually have all these fields. I think only name, email, image, username, createdAt, updatedAt should be enough.

packages/fauna/package.json
```diff
@@ -21,10 +21,19 @@
     "access": "public"
   },
   "scripts": {
-    "test": "exit 0"
+    "reset": "export FAUNA_ADMIN_KEY=secret;export FAUNADB_DOMAIN=localhost; export FAUNADB_SCHEME=http;export FAUNADB_PORT=8443;fauna-schema-migrate apply all",
```
pbteja1998, line 24: This can be cleaned up a bit and make it work in all types of system using cross-env

For reference:
https://github.com/pbteja1998/nextjs-starter/blob/main/package.json#L20

packages/fauna/package.json
```diff
@@ -21,10 +21,19 @@
...
+    "test:db:start": "docker run -d --rm --name faunadb -p 8443:8443 fauna/faunadb",
```
pbteja1998, line 25: Don't know what will happen if I already have a docker with same name running. Should we change the name to something else to avoid conflicts?

balazsorban44: Changed the container name to `next-auth-fauna-test`, thanks :wink:

packages/fauna/package.json
balazsorban44: Thanks, good idea!

balazsorban44:
https://hub.docker.com/r/amazon/dynamodb-local/

ndom91:
Hey I added basic postgres support to the typeorm-legacy tests. The container starts up, gets seeded, but I'm getting some strange type errors when running the basic tests

```
 FAIL  tests/postgresql/index.test.ts
  ● Test suite failed to run

    ../../basic-tests.ts:10:23 - error TS2344: Type 'Adapter<any, any, any, any>' does not satisfy the constraint '(...args: any) => any'.
      Type 'Adapter<any, any, any, any>' provides no match for the signature '(...args: any): any'.

    10   adapter: ReturnType<Adapter>
                             ~~~~~~~
    ../../basic-tests.ts:59:46 - error TS2344: Type 'Adapter<any, any, any, any>' does not satisfy the constraint '(...args: any) => any'.

    59     adapter: (null as unknown) as ReturnType<Adapter>, // TODO: Make it optional on AppOptions
                                                    ~~~~~~~
    ../../basic-tests.ts:70:16 - error TS2314: Generic type 'AdapterInstance<TUser, TProfile, TSession, TVerificationRequest>' requires 4 type argument(s).

    70   let adapter: AdapterInstance<any, any, any>
                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    ../../basic-tests.ts:137:15 - error TS2532: Object is possibly 'undefined'.

    137         now + appOptions.session.maxAge
                      ~~~~~~~~~~~~~~~~~~~~~~~~~
```

Maybe its just something odd in my environment.

balazsorban44:
If I had to guess, it's an older version of next-auth. Could you report back which version is in your node modules?

ndom91:
@balazsorban44 nuked node_modules, reinstalled, and good to go!

image

balazsorban44:
just have to add the tests for verification request and account and I think this could be merged. postponing dynamodb as I have no experience with it whatsoever

balazsorban44:
I added some tests to the verification request flow, and we now have roughly **~73%** code **coverage**/adapter on **average**! :tada: :heavy_check_mark:

balazsorban44:
I decided to do Account tests in a different PR.

Issue #68, opened 2021-04-25
Create stable release of all the adapters


balazsorban44:

ndom91:
Just wanted to drop a comment here for us not to forget to clean up the `typeorm-legacy` / `prisma-legacy` naming stuff.

Maybe drop the two 'legacy' versions and just roll with the "new" `typeorm` and `prisma` implementations once this is ready :+1:

balazsorban44:
yes, we could try that. I'm even hoping that maybe we could have one unified Prisma adapter, it feels overcomplicated to maintain two almost identical versions.

cc @kripod

ndom91:
Yeah that's what I'm trying to get at - just simplify this whole typeorm/prisma two version thing

kripod:
@balazsorban44 the legacy adapter uses a different schema with worse indexing capabilities, so it isn’t compatible with the new version at all.

ndom91:
@kripod so there would be no easy way to migrate users? Even if we supplied a schema migration script?

Maybe we have to keep around the `prisma-legacy` adapter for a while then and keep recommending people move to / only use the "new" version and then after a few months  / major versions slowly deprecate it. What do you guys think?

kripod:
@ndom91 Migration is absolutely possible. At least, we did it without severe outage at [Copyfolio](https://copyfol.io/). Providing a codemod is a great idea and I totally support keeping only the more recent version of the adapter.

balazsorban44:
To be able to get rid of the old one, the new one still has to support all the old features (meaning option to change table names) so we don't break anything. People seem to be interested in this as a feature, and if we decide to move forward with it for other adapters, it only makes sense if we do the same with `prisma` as well.

kripod:
We can’t support older schema versions with the new adapter, as the required multi-column unique constraints aren’t there to be used, and thus, queries referencing those constraints would fail, see:

https://github.com/nextauthjs/adapters/blob/ebd8245aa087cb495bd20c61014abc58c1327131/packages/prisma/src/index.ts#L50-L55

https://github.com/nextauthjs/adapters/blob/ebd8245aa087cb495bd20c61014abc58c1327131/packages/prisma/src/index.ts#L99-L105

https://github.com/nextauthjs/adapters/blob/ebd8245aa087cb495bd20c61014abc58c1327131/packages/prisma/src/index.ts#L168-L187

ndom91:
[...]

PR #106, opened 2021-05-26
docs(adapters): update readme version badges -> canary

ndom91:
## Reasoning 💡

Update README badges to point to latest canary version until #68 is complete.

## Checklist 🧢

- [x] Documentation
- ~[ ] Tests~
- [x] Ready to be merged


README.md
```diff
@@ -19,14 +19,14 @@ This repository is a lerna mono-repo which is home to multiple next-auth adapter
 
 Currently, there are the following adapters ready for use
```
ndom91, line 20: Maybe also change the text here until #68 is complete to explicitly recommend / warn to use the canary versions? @balazsorban44

balazsorban44: Yes, I think we should say something like these are experimental adapters, and we would like feedback on them!

balazsorban44 (changes requested):

README.md
ndom91: Alright, updated! What do you think? 

I tried to make it clear what the plan is regarding the legacy  adapters as well.

balazsorban44 (approved):
Much better!

Issue #125, opened 2021-06-03
DynamoDB Adaptor documentation should be adjusted


manishprivet:
Hi, so I wanted to try next-auth in a serverless vercel app, and we were using DynamoDB, so followed the setup instructions from here https://next-auth.js.org/adapters/dynamodb, but somehow the adaptor was not working. So I decided to inspect the code and found out that the code was completely broken, with wrong sdk calls, and wrong instructions in the documentation itself.

I'll explain it one by one.

# 1. Wrong documentation

In the docs, we're supposed to use the docs like this
```js
adapter: DynamoDBAdapter({
    AWS,
    tableName: "next-auth-test",
  }),
```

But, if we look at the adaptor code
https://github.com/nextauthjs/adapters/blob/4177a25996560287f9a4f08a92ca1c44d7eaf36e/packages/dynamodb/src/index.ts#L5-L13
It's supposed to have two parameters, one for the DynamoDB client, and the other with an object containing the `tableName`.

So, the current documentation should be

```js
adapter: DynamoDBAdapter(
  new AWS.DynamoDB(),
  { tableName: "next-auth-test" })
```

# 2. Wrong use of AWS SDK and DynamoDB Client
Some of the functions in `aws-sdk` DynamoDB client for CRUD operations are:

1. `putItem`
2. `getItem`
3. `updateItem`

While the ones used here is

1. `get`
2. `put`
3. `update`

which simply throws an error like `client.put is not a function`.

# 3. Wrong query syntax

Let's take this example
https://github.com/nextauthjs/adapters/blob/4177a25996560287f9a4f08a92ca1c44d7eaf36e/packages/dynamodb/src/index.ts#L33-L45
This is fine for next-auth as it expects its response in this format only.

But the correct query for DynamoDB should contain

```js
const item: any = {
  pk: {S: `USER#${userId}`},
  sk: {S: `USER#${userId}`},
  id: {S: userId},
  type: {S: "USER"},
  name: {S: profile.name || ""},
  email: {S: profile.email || ""},
  image: {S: profile.image || ""},
  username: {S: profile.username || ""},
  emailVerified:  {S: profile.emailVerified?.toISOString() ?? ""},
  createdAt: {S: now.toISOString()},
  updatedAt: {S: now.toISOString()},
 }
```
So any entity object needs to be parsed between for next-auth and DynamoDB format, in all functions.

Or it'll simply throw an infinite number of validation errors.

[...]

PR #126, opened 2021-06-03
Updated DynamoDB Adaptor documentation

manishprivet:
## Reasoning 💡
#90 changed the way DynamoDB was implemented, and the documentation needed to be updated accordingly.

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## Affected issues 🎟
Addresses #125


balazsorban44 (changes requested):
So in the source I see that table name has a default value, so it is actually not needed for a default setup. It defaults to `"next-auth" `

packages/dynamodb/README.md
```diff
@@ -30,7 +30,7 @@ npm install next-auth @next-auth/dynamodb-adapter@canary
 
 2. Add this adapter to your `pages/api/[...nextauth].js` next-auth configuration object.
 
-You need to pass `aws-sdk` to the adapter in addition to the table name.
+You need to pass `DocumentClient` from `aws-sdk` to the adapter in addition to the table name.
```
balazsorban44, line 33: Same as https://github.com/nextauthjs/next-auth/pull/2125#discussion_r645206437

balazsorban44 (changes requested):

balazsorban44 (changes requested):
I would rather not mention table name this way, just show the bare minimal implementation, and maybe add an options section in the docs, under schema for example.

ndom91:
Looks good to me!

stephankaag:
Hi @manishprivet 

With the current table design sessions cannot be looked up with the session's AccessToken. Is that on purpose? If so: what's the point of the AccessToken?

manishprivet:
> Hi @manishprivet 
> 
> With the current table design sessions cannot be looked up with the session's AccessToken. Is that on purpose? If so: what's the point of the AccessToken?

I didn't work on the table design actually. I think @PolMrt could answer your question

PR #127, opened 2021-06-04
Update README - Package Table

ndom91:
## Reasoning 💡

Added a column with badges for the # of weekly downloads for each adapter package. What do you guys think? 

**Example:**

> image

Maybe instead of downloads the column should be the bundlephobia package size badge? 

What do you guys find more useful?

## Checklist 🧢

- [x] Documentation
- ~[ ] Tests~
- [x] Ready to be merged


balazsorban44:
Not sure these are extremely interesting metrics, but I don't oppose to them either.

ndom91:
Yeah your probably right. The latest version, docs, and npm pkg link make a lot of sense - anything else useful?

PR #132, opened 2021-06-14
feat(prisma): pass all profile parameters when creating prisma user

### LINKED ISSUE #131: Custom fields with prisma adapter
I want to save the username of the user's profile on the GitHub OAuth, but saving custom fields for prisma does not seem to be documented anywhere. I have tried to do this:
```ts
export default NextAuth({
  providers: [
    Providers.GitHub({
      clientId: envConfig.get('gitHub.clientId'),
      clientSecret: envConfig.get('gitHub.clientSecret'),
      profile(profile) {
        return {
          // @ts-expect-error id does exist
          id: profile.id.toString(),
          // USERNAME DEFINED HERE
          username: profile.login,
          bio: profile.bio,
          name: profile.name ?? profile.login,
          email: profile.email,
          image: profile.avatar_url,
        } as UserWithStringId
      },
    }),
  ],
  adapter: PrismaAdapter(prisma),
  session: { jwt: true },
})
```
but `username` is not added to `data` when `prisma.user.create` is called and the function errors. Is there any way to do this?

ndom91:
## Reasoning 💡

Allow creating custom fields in our User table by passing all profile fields to `prisma.user.create()`.

I did some googling and couldn't find much in terms of info on schema with dynamic fields, etc. maybe @williamluke4 can chime in here - is this okay to do in prisma?

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## Affected issues 🎟

Closes #131


sachinraja:
Anything I can do to help this get merged? Dynamic fields should be fine as long as you've defined them in your schema.

balazsorban44:
thank you @sachinraja. See the description of this PR. Could you answer the concerns there?

ndom91:
> Anything I can do to help this get merged? Dynamic fields should be fine as long as you've defined them in your schema.

So in this PR we've changed nothing about the schema. This sounds more like something each user would need to do, right? Or can we define a dynamic name for the field in the schema as well?

sachinraja:
@ndom91This is definitely something the user would handle by changing their schema. Although it may be a breaking change because Prisma will throw an error if you pass a value for a field that is not defined in the schema. The user may also end up passing unwanted field values, as the current system only picks a few fields out of the passed object.

ndom91:
Okay understand, so we'd have to document it well that the user must define the fields in their prisma schema file and pass those and only those extra fields through the profile object, etc. Am I missing anything?

sachinraja:
No, that should be all.

ndom91:
Alright I added some info in the README - if theres nothing missing I'll update the docs page as well after this gets merged.

@balazsorban44 what do you think?

sachinraja:
Should all parameters also be passed for [`updateUser`](https://github.com/nextauthjs/adapters/blob/canary/packages/prisma/src/index.ts#L59)?

balazsorban44:
Sounds good to me, @ndom91.


@sachinraja here it is https://github.com/nextauthjs/next-auth/blob/86baefdd9df9246669d4f788aaab4774605c256a/src/server/lib/callback-handler.js#L106

sachinraja:
Thanks @balazsorban44. It looks to be an internal function then. If that's the only place it's used then, the current implementation is fine.

packages/prisma/README.md
```diff
@@ -51,6 +51,29 @@ export default NextAuth({
...
+  createdAt     DateTime  @default(now())
+  updatedAt     DateTime  @updatedAt
+  accounts      Account[]
+  sessions      Session[]
+}
+```
+
+When running `prisma migrate`, Prisma will create the User table with the above schema in your specified underlying database.
+
+Then, during the first sign-in NextAuth.js will return the simpler prototype user object, but after the first signin, NextAuth.js will return the full custom User object, including your custom field and userId, in callbacks such as [`signIn`](https://next-auth.js.org/configuration/callbacks#sign-in-callback).
```
sachinraja, line 75: First use of sign-in is "sign-in", second use is "signin".

ndom91: One is referring to the act of signing in and one to the specific callback. You think they should both be `signIn`, for example?

balazsorban44: I would go with sign-in. Let's don't confuse function names and the action of signing in.

packages/prisma/README.md
```diff
@@ -51,6 +51,29 @@ export default NextAuth({
...
+  createdAt     DateTime  @default(now())
+  updatedAt     DateTime  @updatedAt
+  accounts      Account[]
+  sessions      Session[]
+}
+```
+
+When running `prisma migrate`, Prisma will create the User table with the above schema in your specified underlying database.
+
+Then, during the first sign-in NextAuth.js will return the simpler prototype user object, but after the first signin, NextAuth.js will return the full custom User object, including your custom field and userId, in callbacks such as [sign-in](https://next-auth.js.org/configuration/callbacks#sign-in-callback).
```
sachinraja, line 75:
```suggestion
Then, during the first sign-in NextAuth.js will return the simpler prototype user object, but after the first sign-in, NextAuth.js will return the full custom User object, including your custom field and userId, in callbacks such as [`signIn`](https://next-auth.js.org/configuration/callbacks#sign-in-callback).
```

anglinb:
👋 You all might be aware of this behavior but it was surprising to me. By passing all the profile fields, the primary key on the users table will now default to the `profile.id` rather than whatever the database would generate. If you have multiple providers then your primary key could be (1) very different length and (2) more prone to collisions based on which account you signed in with first.

When I signed in with GitHub, for example, my primary key in the Users table was `2637602` instead of something like `ckqeycrjc00154onepq2yf6xs` which was created by letting Prisma just assign a value with `@default(cuid())`.

If this is expected, feels like a breaking change worth documenting. Otherwise, pretty easy fix to achieve what you're after without changing how the id field works. I've wrapped the prisma adapter so I could tweak a couple other things and undid this change by doing this: 

```
let { id, ...restOfProfile } = profile;
return prisma.user.create({
  data: {
    name: profile.name,
    email: profile.email,
    image: profile.image,
    emailVerified: profile.emailVerified?.toISOString() ?? null,
     ...restOfProfile,
    personalTeamId: personalTeam.id,
  },
})
```

ndom91:
Hey thanks a lot for the heads up! 

So the provider `id` is overwriting our default primary key `id` field due to this spreading in all of their fields, right? In your example it looks like you're excluding the provider `id` and not writing that into the db. @balazsorban44, is it important for us to have the provider `id`? i.e. the github user id? What if we save that in a field called `providerId` or something along those lines?

sachinraja:
I think it would be better to handle custom fields in a different way as the profile object can still have many names that clash with the user's schema fields. Is there a way we can tell if the user passed in a custom profile object?

sachinraja:
Or what if there was a special property on the object like this:
```js
profile(profile: Profile & GitHubFields) {
  return {
    name: profile.name ?? profile.login,
    email: profile.email,
    image: profile.avatar_url,
    __nextAuthCustomFields: {
      id: profile.id.toString(),
      username: profile.login,
      bio: profile.bio,
    },
  }
}
```

ndom91:
Hmm I like this approach, because we can never be sure what fields every provider is passing and what they're called, etc. By putting them into their own special properties object we've effectively namespaced the custom profile fields :+1:

ndom91:
Opened a separate issue for this discussion here: https://github.com/nextauthjs/adapters/issues/151

PR #136, opened 2021-06-15
fix(prisma): ensure that `providerAccountId` is a string

williamluke4:


conquestmakes:
@williamluke4 the problem also occurs in the linkAccount method so you'll want to fix it there too

balazsorban44:
So I will ignore this PR for now, as in the next version of `next-auth`, we have made sure that this is passed as a string (https://github.com/nextauthjs/next-auth/pull/2108)

PR #140, opened 2021-06-23
feat(dynamodb): store expires as a timestamp in Unix epoch time format

stephankaag:
To cleanup expired sessions it would be great to take advantage of Dynamo's built-in TTL. 

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/time-to-live-ttl-before-you-start.html

This feature requires the expires attribute to a Number.


stephankaag:
Note: this change might conflict with existing sessions.

balazsorban44:
Note that adapters have changed significantly in v4, so this adapter has to be rewritten from scratch.
See #171, #182, #183, #185, #186, #193 for inspiration

PR #150, opened 2021-06-27
build(ci): create release notes

balazsorban44:
Currently, we are using CHANGELOG.md files to create release notes. To better match our main repo, I made some changes to instead push these changes as release notes.

I gathered some info, and `lerna` does support such behavior.

See:
1. https://github.com/lerna/lerna/tree/main/commands/publish#options
> lerna publish supports all of the options provided by lerna version in addition to the following:

2. https://github.com/lerna/lerna/tree/main/commands/version#--create-release-type
> When run with this flag, lerna version will create an official GitHub or GitLab release based on the changed packages. Requires --conventional-commits to be passed so that changelogs can be generated.
GH_TOKEN (required) - Your GitHub authentication token (under Settings > Developer settings > Personal access tokens).

We also named our `GH_TOKEN` as `GITHUB_TOKEN` which is not the assumed format by `lerna`. I renamed it.

`GITHUB_TOKEN` is a value that comes from organization secrets. I asked @iaincollins to provide me with access to be able to have more control later on, but it will be his call if he trusts me with that kind of responsibility.

(In the core repo, `semantic-release` uses that to create release notes. See https://github.com/nextauthjs/next-auth/blob/ad6c13cdc901e6da9c4aa06535e322429a8b1562/.github/workflows/release.yml#L47)


ndom91 (approved):
LGTM :+1:

PR #162, opened 2021-07-02
test(fauna): increase wait until DB is online

ndom91:
## Reasoning 💡

Test were a bit flaky due to DB not being online within 10sec in some cases. Increased sleep to 20 seconds to ensure db is online and tests can be run against it.

## Checklist 🧢

- ~[ ] Documentation~
- [x] Tests
- [ ] Ready to be merged


packages/fauna/tests/test.sh
```diff
@@ -14,7 +14,7 @@ docker run -d --rm \
 fauna/faunadb
 
 echo "Waiting 10 sec for db to start..."
```
balazsorban44, line 16: should update this echo command

balazsorban44 (changes requested):

packages/fauna/tests/test.sh
ndom91: Ah yes, good call!

balazsorban44 (approved):
LGTM

PR #171, opened 2021-07-15
feat(prisma): use new Adapter API

balazsorban44:
Based on https://github.com/nextauthjs/next-auth/pull/2361

This PR also removes the `prisma-legacy` adapter entirely, so we can focus on a single one moving forward. Having two versions caused a fair amount of confusion anyway.


packages/prisma/prisma/schema.prisma
```diff
@@ -2,33 +2,38 @@ datasource db {
...
-  refreshToken       String?
-  accessToken        String?
-  accessTokenExpires DateTime?
-  createdAt          DateTime  @default(now())
-  updatedAt          DateTime  @updatedAt
-  user               User      @relation(fields: [userId], references: [id])
-  @@unique([providerId, providerAccountId])
+  type               String
+  provider           String
+  refresh_token      String?
```
sastraxi, line 15: Nitpick: is there a reason why some fields are in snake_case now? Prisma's docs [use camelCase for their examples](https://www.prisma.io/docs/about/about-the-docs#naming-conventions-for-tables-and-columns), and this seems to be true of most Prisma codebases I've seen.

balazsorban44: Mostly because the OAuth spec defines these as snake_case, and I thought it would make things easier if we just forward them "as-is".

balazsorban44:
TODO: Update test suite

ndom91:
Hey I mentioned I was doing some work with prisma myself, and I finally ran into the usecase of the `@@unique([field1, field2])`, like we have for the verification token. 

I think that's only for defining fields to lookup by, that doesn't do anything in the actual db schema to make a union primary key or anything like that. 

I think in prisma we might be good in terms of that security issue, as the table / model doesn't have any other keys to lookup by. So it has to be queried by this combination of fields. 

But definitely something to keep in mind for other adapters, this will probably need an additional custom union primary key of the two fields and probably some logic in the adapter if I understood you and the security issue correctly

ndom91 (approved):
LGTM :+1:

PR #182, opened 2021-08-16
feat(fauna): use new Adapter API

balazsorban44:
See #171, https://github.com/nextauthjs/next-auth/pull/2361

~TODO: Figure out how to cascade deletion of accounts and sessions:~

~In the future, we would like to implement a "delete user" feature in NextAuth. We want to use the `deleteUser` adapter method for that. In addition to removing the actual User, we should remove any remaining user objects like the connected accounts and active sessions. I need help to figure out how to do that.~

~cc @pbteja1998 Could you provide some guidance?~

~Docs says:~
~> There is no cascading delete for user-defined documents. If you have a document "A" that includes a reference to another document "B", deleting "A" does not affect "B", and deleting "B" does not affect "A". You would have to write your own query logic to automatically follow references to perform a cascading delete.~

~https://docs.fauna.com/fauna/current/api/fql/functions/delete?lang=javascript~


packages/fauna/fauna/resources/indexes/user_by_provider_account_id.fql
```diff
@@ -0,0 +1,9 @@
+CreateIndex({
+  name: "user_by_provider_account_id",
```
pbteja1998, line 2: It would be better to name this `account_by_provider_and_provider_account_id`

packages/fauna/fauna/resources/indexes/user_by_provider_and_provider_account_id.fql
```diff
@@ -0,0 +1,9 @@
+CreateIndex({
+  name: "user_by_provider_and_provider_account_id",
+  source: Collection("accounts"),
```
pbteja1998, line 3: Since the `source` is `accounts` collection, may be we should name it `account_by_provider_and_provider_account_id`?

packages/fauna/src/index.ts
```diff
@@ -0,0 +1,216 @@
...
+              Lambda("ref", Delete(Var("ref")))
+            ),
+            deleteAccounts: Map(
+              Paginate(Match(indexes.AccountsByUser, userId)),
+              Lambda("ref", Delete(Var("ref")))
+            ),
+            user: Delete(Ref(Users, userId)),
+          },
+          true
+        )
```
pbteja1998, line 141: I think this is fine, but since we don't need to care about the return value here, we can maybe use `Do` and `ForEach` instead of `Let` and `Map`. This is just a suggestion. What you already have is fine too.

packages/fauna/fauna/resources/indexes/user_by_provider_account_id.fql
balazsorban44: Changed, thanks!

packages/fauna/fauna/resources/indexes/user_by_provider_and_provider_account_id.fql
balazsorban44: True. It IS used to retrieve the actual User in the adapter method, but here it should be renamed. Will change!

packages/fauna/src/index.ts
balazsorban44: Could you leave a code suggestion?

pbteja1998:
```suggestion
        Do(
            Foreach(
              Paginate(Match(indexes.SessionsByUser, userId)),
              Lambda("ref", Delete(Var("ref")))
            ),
            Foreach(
              Paginate(Match(indexes.AccountsByUser, userId)),
              Lambda("ref", Delete(Var("ref")))
            ),
            Delete(Ref(Users, userId))
        )
```

packages/fauna/fauna/resources/indexes/user_by_provider_and_provider_account_id.fql
balazsorban44: Changed!

pbteja1998 (approved):

ndom91 (approved):
LGTM as well, man is fauna less inuitive though than prisma imo :sweat_smile: the prisma methods and object arguments "just make sense", where fauna seems super cryptic at times

Tests all passed, etc :tada: I had some trouble initially though - had to delete my `node_modules` and `package-lock.json` a few times and remember to use node 16+ (`randomUUID`) and then finally with node 16 and switching to yarn I was able to get it to run and all pass!!

Issue #180, opened 2021-08-11
Add support for firebase auth


HawKyre:
Hello, I wondered if it could be possible to add support for firebase auth in the firebase adapter, since it's also a part of the firebase ecosystem and right now everything is being handled by firestore.

balazsorban44:
I don't see any API in Firebase Auth that makes it possible to create users programmatically. Am I wrong?

balazsorban44:
Correction, I was reminded that there *is* an API! https://firebase.google.com/docs/auth/admin/manage-users I will look into it then!

balazsorban44:
I would be happy if anyone interested in this could have a look as well. I am happy to give support.

srijans38:
@balazsorban44  I'm interested in this. Just took a look. Found two issues with this so far:
1.  To integrate with Firebase Auth, we'd have to use "firebase-admin" which is a totally separate package from "firebase" web client and requires different setup.
2. What if someone wants to just use firestore as a db without firebase auth.

PR #183, opened 2021-08-16
feat(firebase): use new Adapter API

balazsorban44:
See nextauthjs/adapters#171, nextauthjs/adapters#182

Firebase introduces a more modular version from 9 and up. The adapter has been rewritten to work with v8. I will see if I can make the upgrade:

https://firebase.google.com/docs/web/learn-more#v9-new-apps

A few things to figure out before we can merge this:
- should we support Firebase 8 or Firebase 9 (or both?) nextauthjs/next-auth#3827 
- can we integrate with Firebase Auth nextauthjs/adapters#180

Any help with these would be appreciated! :pray:


packages/firebase/src/index.ts
```diff
@@ -1,283 +1,243 @@
...
+  Accounts: "accounts",
+  VerificationTokens: "verificationTokens",
+} as const
+
+export const format: FirestoreDataConverter<any> = {
+  toFirestore(object) {
+    const newObjectobject: any = {}
+    for (const key in object) {
+      const value = object[key]
+      if (value === undefined) continue
```
chanceaclark, line 39: 👋 I've been playing around with this branch on a personal project and noticed a small issues when using the `GoogleProvider` for authentication. This line specifically checks for `undefined`, but in the case of a response from Google where `emailVerified` === `null`, reading the value below in `fromFirestore` on `value.toDate` will throw:
```
https://next-auth.js.org/errors#session_error Cannot read property 'toDate' of null TypeError: Cannot read property 'toDate' of null
```

Not sure what the best approach would be here but maybe you either want to:
1. Exclude `null` values from when writing to firebase (probably not ideal and not recommended since `null` is a valid datatype) or...
2. Ensuring `value` is defined below, before calling `toDate()`. I'll leave a suggestion below, or..
3. All of above 🤷 Again, maybe not ideal since `null` is a valid datatype but you have more context than me 😅

packages/firebase/src/index.ts
```diff
@@ -1,283 +1,243 @@
...
+      newObjectobject[key] = value
+    }
+    return newObjectobject
+  },
+  fromFirestore(snapshot) {
+    if (!snapshot.exists()) return null
+    const newUser: any = { ...snapshot.data(), id: snapshot.id }
+    for (const key in newUser) {
+      const value = newUser[key]
+      if (value.toDate) newUser[key] = value.toDate()
```
chanceaclark, line 49:
```suggestion
      if (value && value.toDate) newUser[key] = value.toDate()
```
or even better
```suggestion
      if (value?.toDate) newUser[key] = value.toDate()
```

amesas:
Hi!

Is this pull request in a usable state?

packages/firebase/src/index.ts
ceevie: had the same issue with firebase v9. this fix works great.

saseungmin: Thank you for your suggestion! Awesome!

SahilAujla:
Does anyone know how to use @next-auth/firebase-adapter with NextAuth.js version 4?

srijans38:
@balazsorban44  Please see https://github.com/nextauthjs/adapters/issues/180#issuecomment-1004126562. The issue has been closed by stale bot. I also have firebase auth working here #363. Need some feedback to see what more can be done.

PabloFX:
@balazsorban44 is there any chance to upgrade firebase adapter in the near future?

packages/firebase/src/index.ts
JimmyCHL: yes, this solve the issue and it work in localhost. However, when i tried to deploy to Vercel, it could be build successfully, but it would have problem when login with google. When login with the google, it always show callback error, but in the localhost environmnet, there is no issue. is threre anything i should concern?

balazsorban44:
This repository has been merged into a monorepo. Follow-up PR is here: https://github.com/nextauthjs/next-auth/pull/3873

PR #186, opened 2021-08-20
feat(neo4j): use new Adapter API

EarthlingDavey:
See #171, nextauthjs/next-auth#2361


EarthlingDavey:
Thanks for starting the tests. If the error is something that I can fix, please let me know :)

.github/workflows/release.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - name: Test (changed packages)
         run: lerna run test --concurrency 3 --since HEAD~
       - name: Build (changed packages)
-        run: lerna run build --concurrency 3 --since HEAD~
```
balazsorban44, line 33: Not sure where this change come from, see `next`:

https://github.com/nextauthjs/adapters/blob/5c915823c43867dd54adf3977c26968baa5ec11f/.github/workflows/release.yml#L33

Could you maybe update your branch by merging `next` into it?

packages/neo4j/README.md
```diff
@@ -0,0 +1,63 @@
...
+  providers: [
+    Providers.Google({
+      clientId: process.env.GOOGLE_ID,
+      clientSecret: process.env.GOOGLE_SECRET,
+    }),
+  ],
+  adapter: Neo4jAdapter(neo4jSession),
+  // ...
+})
+```
```
balazsorban44, line 55:
```suggestion
// Setup your neo4j driver instance
const driver = neo4j.driver(
  "bolt://localhost",
  neo4j.auth.basic("neo4j", "password")
).session()
const neo4jSession = driver.session()

export default NextAuth({
  adapter: Neo4jAdapter(neo4jSession),
  // Rest of your NextAuth config
})
```

Let's keep this minimal. (Would like to transform the other examples to this gradually. The example code should only focus on the adapter-related stuff IMO.)

packages/neo4j/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "oauth",
+    "neo4j"
+  ],
+  "private": false,
+  "publishConfig": {
+    "access": "public"
+  },
+  "scripts": {
+    "test:watch": "./tests/test.sh -w",
+    "test": "yarn build && ./tests/test.sh",
```
balazsorban44, line 25:
```suggestion
    "test": "./tests/test.sh",
```

Build is already run in the pipeline, no need here.

packages/neo4j/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "peerDependencies": {
+    "next-auth": ">=4"
+  },
+  "devDependencies": {
+    "@types/uuid": "^8.3.0"
+  },
+  "dependencies": {
+    "neo4j-driver": "^4.3.0",
+    "uuid": "^8.3.2"
+  }
```
balazsorban44, line 41:
```suggestion
  "peerDependencies": {
    "neo4j-driver": "^4.3.0",
    "next-auth": ">=4",
  },
  "devDependencies": {
    "@types/uuid": "^8.3.0",
    "neo4j-driver": "^4.3.0",
  },
  "dependencies": {
    "uuid": "^8.3.2"
  }
```

Let the user use their own version of `neo4j-driver` by adding it to `peerDependencies`. Other than that, we only use the package for types, so it can be in `devDependencies` as well.

packages/neo4j/src/account.ts
```diff
@@ -0,0 +1,96 @@
...
+import type { Account } from "next-auth"
+import { v4 as uuid } from "uuid"
+
+export const linkAccount = async (
+  neo4jSession: typeof neo4j.Session,
+  data: Account
+) => {
+  const { userId, providerAccountId, provider, ...accountData } = data
+
+  // TODO: Question, account.expires is an int. here
```
balazsorban44, line 11: As far as I know, there is no `expires` on `Account`. You might be thinking `expires_at`, which comes from here: https://github.com/panva/node-openid-client/blob/772306cc82b72ace9361f4a945249a5c35a8f921/types/index.d.ts#L714-L718

Since we just forward these values, we will respect their type as well, so this is kept as a number.

packages/neo4j/src/account.ts
```diff
@@ -0,0 +1,96 @@
...
+        a.id = $id
+      SET 
+        a += $accountData
+
+      MERGE (u)-[:HAS_ACCOUNT]->(a)
+
+      RETURN a, u.id AS userId
+      `,
+        {
+          id: uuid(),
```
balazsorban44, line 36: Would be awesome, if there was a way to generate an `id` upon creation in `neo4j`. We could get rid of the `uuid` dependency. Do you think it is possible? Most of the other adapters just create an `id` automatically, without you needing to pass it here.

packages/neo4j/src/index.ts
```diff
@@ -0,0 +1,97 @@
...
+    async createSession(data) {
+      return await createSession(neo4jSession, data)
+    },
+
+    async getSessionAndUser(sessionToken) {
+      return await getSessionAndUser(neo4jSession, sessionToken)
+    },
+
+    // TODO:
+    // Question: force parameter is no longer used?
```
balazsorban44, line 79: No. Most of the logic is moved to the core, but the `force` param wasn't passed anywhere, so technically it has **never** been used... :shrug:

balazsorban44 (changes requested):
Thank you! I just reviewed this, and it looks like a good start! I had some comments, to begin with.

One unfortunate thing is that because of some GitHub security measures, it looks like we cannot run tests on PRs from fork repositories with our current flow. (cc @ndom91 we should work on that. The issue is that the `GITHUB_TOKEN` secret is not passed to forks PRs. We could create a minimal job that only runs on these types of PRs and just run tests, not publishing.)

Running the tests locally, everything passes, so congrats on that! 🎉 looks like you nailed the logic.

Although I have some general advice for reorganizing the code. I see a lot of repetitive stuff in different files, and most of them boil down to the same things:

1. try..catch wrapper/log,
2. `neo4jSession`
3. Data mutation `from` and `to` neo4j

I recommend that you rather create an abstraction that handles all these, and instead of having separate files, you could easily fit everything in a single one!

EarthlingDavey:
@balazsorban44 thanks for your feedback, and answers to my questions... I'll work my way through things

.github/workflows/release.yml
EarthlingDavey: It's odd, I already merged from next. And merged agin and this change isn't in the merge. 

I don't know why, so I made the edit manually.

packages/neo4j/package.json
EarthlingDavey: That makes sense. Fixed: https://github.com/usedandloved/nextauthjs-adapters/commit/4f6eef42c007b71ff7a08af2980473ef8b2af7c0

packages/neo4j/src/account.ts
EarthlingDavey: I'll look into this and speak to the neo4j community. 

My reason for not using internal ids, [neo4j docs](https://neo4j.com/docs/cypher-manual/current/clauses/match/#get-node-rel-by-id):

> Neo4j reuses its internal ids when nodes and relationships are deleted. This means that applications using, and relying on internal Neo4j ids, are brittle or at risk of making mistakes. It is therefore recommended to rather use application-generated ids.

Alternative is to use [Neo4j APOC Library](https://neo4j.com/developer/neo4j-apoc/) to move the dependency from from the app to the database. But, this may just be a different inconvenience for developers. @ndom91 and I had a chat about this and decided to go with the js package for uuid because it was 'easiest' for the user.

packages/neo4j/src/index.ts
```diff
@@ -0,0 +1,258 @@
...
+              undefined !== emailVerified
+                ? `, u.emailVerified = datetime($emailVerified)`
+                : ``
+            }
+          RETURN properties(u) AS u`,
+        {
+          userData: { id: uuid(), ...userData },
+          emailVerified,
+        }
+      )
```
balazsorban44, line 29:
```suggestion
export function Neo4jAdapter(neo4jSession: typeof neo4j.Session): Adapter {
  const query = neo4jWrap(neo4jSession)
  return {

    async createUser({ emailVerified, ...userData }) {
      return await query(
        ` CREATE (u:User)
          SET
            u += $userData
            ${
              undefined !== emailVerified
                ? `, u.emailVerified = datetime($emailVerified)`
                : ``
            }
          RETURN properties(u) AS u`,
        {
          userData: { id: uuid(), ...userData },
          emailVerified,
        }
      )
```

packages/neo4j/src/index.ts
```diff
@@ -0,0 +1,258 @@
...
+          RETURN properties AS deletedUser`,
+        {
+          id,
+        }
+      )
+    },
+
+    // * * * * * * * * * * * * * * * *
+    // Account
+    // * * * * * * * * * * * * * * * *
```
balazsorban44, line 110:
```suggestion
```

packages/neo4j/src/index.ts
```diff
@@ -0,0 +1,258 @@
...
+        ` MATCH (u:User)-[:HAS_ACCOUNT]->(a:Account { 
+            providerAccountId: $providerAccountId, 
+            provider: $provider 
+          })
+          WITH u, a, properties(a) AS properties
+          DETACH DELETE a 
+          RETURN properties { .*, userId: u.id } AS deletedAccount`,
+        {
+          ...provider_providerAccountId,
+        }
```
balazsorban44, line 148:
```suggestion
        provider_providerAccountId
```

packages/neo4j/src/index.ts
```diff
@@ -0,0 +1,258 @@
...
+            a += $accountData
+          MERGE (u)-[:HAS_ACCOUNT]->(a)
+          RETURN a { .*, userId: u.id } AS a`,
+        {
+          id: uuid(),
+          userId,
+          providerAccountId,
+          provider,
+          accountData,
+        }
```
balazsorban44, line 132: Technically, everything here is `accountData`, could you explain the reason for this property, rather than using the spread operator?

packages/neo4j/src/index.ts
```diff
@@ -0,0 +1,258 @@
...
+          RETURN properties { .*, userId: u.id } AS deletedAccount`,
+        {
+          ...provider_providerAccountId,
+        }
+      )
+    },
+
+    // * * * * * * * * * * * * * * * *
+    // Session
+    // * * * * * * * * * * * * * * * *
```
balazsorban44, line 154:
```suggestion
```

packages/neo4j/src/index.ts
```diff
@@ -0,0 +1,258 @@
...
+          RETURN properties { .*, userId: u.id } AS deletedSession`,
+        {
+          sessionToken,
+        }
+      )
+    },
+
+    // * * * * * * * * * * * * * * * *
+    // VerificationToken
+    // * * * * * * * * * * * * * * * *
```
balazsorban44, line 229:
```suggestion
```

balazsorban44 (changes requested):
Thank you! This looks much better! I think it can be even further simplified, with a small API change. I left a suggestion for the `createUser` method, but you can follow it for the other methods as well. Basically, your wrapper should only take a single argument (`neo4jSession`) and return a function that you pass your queries and data to. This will make this API super clean, in my opinion!

packages/neo4j/src/index.ts
EarthlingDavey: I see what you mean but there is a reason.
* On this line I am setting account data to be everything except for userId, providerAccountId, provider.
  `async linkAccount({ userId, providerAccountId, provider, ...accountData }) {`

* Finding or creating the Account node.
  ```
  MERGE (a:Account { 
    providerAccountId: $providerAccountId, 
    provider: $provider 
  }) 
  ```

* Setting the account data here
  `SET a += $accountData`
  userId should not be set on this node because a relationship to a user node is created
  and, it is unnecessary to set providerAccountId and provider properties again.

packages/neo4j/src/index.ts
EarthlingDavey: Good suggestion. I've done this. But, I've run out of time so some of the typescript types are set to any, where I guess they could be more specific.

packages/neo4j/src/account.ts
EarthlingDavey: @balazsorban44 do you want to have a chat about this? No-one has got back to me in the neo4j community chat. 
I feel like the `uuid` is an acceptable compromise. I don't have much more time to spend on this, so if you're not sure, then let's arrange a quick chat to work something out. Cheers.

balazsorban44: Thanks for checking, no problem! We can let this be for now, and handle it a future optimization opportunity.

balazsorban44:
hi, I'm sorry for the delay here. I'm currently focusing elsewhere. @ndom91, @lluia would you have time to help here?

nevertheless, nothing wrong with using this code directly i  your project, only better, you will have real world testing for us... 😅

EarthlingDavey:
@balazsorban44 thanks for getting back to me. I understand ... and I do not want to pile any additional pressure on ... haha developers usually have a lot of other stresses!

I appreciate you maybe delegating to a different contributor. I hope they can help. And, btw I'll try and be responsive to questions, to make their life easier :D

Just so you can understand my POV... I learned TS for this PR, I re-wrote my work to the new API, and responded to one round of reviews.

I'm fine to use my work directly. But, it would be nice for me to wrap this up as I already do plenty context switching - being a solo full stack startup dev, with an imminent launch 😅 

I will of course fix any noe4j issues going forward (I did already fix one bug report).

Cheers all, and have a good day ✌🏼

ndom91:
Hi everyone! Just approved the latest changes to run through the tests. Sorry I haven't kept up entirely here. Whats the latest status? Is it more or less ready to go from your point of view @EarthlingDavey ?

EarthlingDavey:
Hi @ndom91 . Thanks for taking a look at this. 

It's passing tests locally and I have capacity to respond and maintain it. So, I think it's ok for you guys to run workflows and if all is ok then merge in :)

balazsorban44:
Before merging, we need a docs page at

https://github.com/nextauthjs/docs/tree/main/docs/adapters

and  a mention of the Adapter on https://github.com/nextauthjs/docs/blob/main/docs/adapters/overview.md

packages/neo4j/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "test": "./tests/test.sh",
+    "build": "tsc"
+  },
+  "files": [
+    "README.md",
+    "dist"
+  ],
+  "peerDependencies": {
+    "neo4j-driver": "^4.0.0",
+    "next-auth": ">=4"
```
balazsorban44, line 34:
```suggestion
    "next-auth": ">4 || 4.0.0-beta.1 - 4.0.0-beta.x",
```

As per #274

balazsorban44, line 34:
```suggestion
    "next-auth": ">4 || 4.0.0-beta.1 - 4.0.0-beta.x",
```

As per #274

balazsorban44 (changes requested):
Could you also add neo4j to the labeler at https://github.com/nextauthjs/adapters/blob/18afbbe3a6fb1c540cc54568378079eb3b101a05/.github/labeler.yml

Can you bring your branch up-to-date with the `next` branch?

packages/neo4j/package.json
balazsorban44: @EarthlingDavey if you fix this, I think we can merge it as is, but we will probably do some cleanup on the code if that is fine by you! Thank you 🙏

EarthlingDavey: @balazsorban44 sure, I'll do this over the weekend 👍 And of course, I would appreciate the cleanup. If you tag me I'll read through the changes to learn from them. Cheers, ha det - have a good weekend

balazsorban44: Never mind,  the suggested change is really small, and I can just merge in as-is and get to work on the cleanup stuff. I would like to get everything ready for the v4 release in time, which is in less than 2 weeks. :slightly_smiling_face:

EarthlingDavey: This is awesome. Thanks man!

balazsorban44:
Running tests locally yields this:

```sh
/var/lib/neo4j/resources/constraints.cypher (No such file or directory)
node:events:368
      throw er; // Unhandled 'error' event
      ^

Error: EACCES: permission denied, scandir '/home/balazs/projects/nextauthjs/adapters/packages/neo4j/tests/resources'
Emitted 'error' event on NodeWatcher instance at:
    at NodeWatcher.checkedEmitError (/home/balazs/projects/nextauthjs/adapters/node_modules/jest-haste-map/build/watchers/NodeWatcher.js:141:12)
    at Walker.emit (node:events:390:28)
    at /home/balazs/projects/nextauthjs/adapters/node_modules/walker/lib/walker.js:63:18
    at FSReqCallback.oncomplete (node:fs:188:23) {
  errno: -13,
  code: 'EACCES',
  syscall: 'scandir',
  path: '/home/balazs/projects/nextauthjs/adapters/packages/neo4j/tests/resources'
}
```

Any ideas @EarthlingDavey?

EarthlingDavey:
@balazsorban44 

I'm not sure. I'll take a look now. Are you on macOS?

balazsorban44:
Linux

EarthlingDavey:
ok.. for now it's safe to remove the following lines from test.sh

```
docker exec \
  -i "${CONTAINER_NAME}" \
  sh -c 'exec cypher-shell -u ${TEST_NEO4J_USER} -p ${TEST_NEO4J_PASS} -f /var/lib/neo4j/resources/constraints.cypher'
```

And perhaps it could be removed permenantly. But first, I will debug on linux now.

Are you running ./test.sh directly from the `nextauthjs/adapters/packages/neo4j/tests` directory?

balazsorban44:
using `yarn test`

EarthlingDavey:
I think I know what is happening, though I haven't had chance to install yarn etc. on a linux box yet.

Because it is being run from one folder up, the variable SCRIPT_DIR is not what I expected.

Could you try sticking `echo $SCRIPT_DIR` on line 4 of `test.sh`
and maybe 

`-v=$SCRIPT_DIR/resources:/var/lib/neo4j/resources \`
should be
`-v=$SCRIPT_DIR/tests/resources:/var/lib/neo4j/resources \`

I'll debug asap and try and find a nice solution.

balazsorban44:
will do, later on. what's that Cypher file BTW? it also stops me from switching branches. I have to delete it with sudo, and then revert the change. should it be gitignored?

EarthlingDavey:
I don't know why, it's nothing special, just a text file. I guess it's the equivelent of a .mysql file etc. 

It's for running once on the neo4j db at startup to add constraints or indexes.

A bunch are commented out because I wrote them, then realised neo4j enterprise and community edition support different contstraints. Hence why most are commented out and I detail/reccomend which ones developers should use in the readme.

100% the tests will run if you delete everything to do with the .cypher file. I suppose it is something I can add back in later after x -platform testing.

EarthlingDavey:
>  I have to delete it with sudo, and then revert the change. should it be gitignored?

I'll try and work out why. It shouldn't be gitignored.

EarthlingDavey:
@balazsorban44 It works ok on my Bodhi linux box. 

- Fresh clone of this repo, 
- checkout next
- cd packages/neo4j
- yarn install
- yarn test

It's also ok on my mac and CI. This makes me think that somehow the permissions got messed up on the .cypher file for you. And maybe a delete and re-clone would help. Sorry I can't give a definite answer.

balazsorban44:
no problem, will check this out later tonight. thanks!

PR #188, opened 2021-08-21
feat(supabase): add supabase Adapter

balazsorban44:
Add official `@next-auth/supabase-adapter` to our adapters list 🎉

Check out: https://supabase.io


balazsorban44:
Figure out if we can use the auth system of supabase instead.

KushibikiMashu:
Great feature !

webdevbry:
would be cool to have a supabase adapter!

DarrylBrooks97:
Wow I need this !

packages/supabase/supabase/schema.sql
```diff
@@ -0,0 +1,35 @@
...
+  id_token VARCHAR(255),
+  session_state VARCHAR(255),
+  CONSTRAINT fk_user FOREIGN KEY("userId") REFERENCES users(id) ON DELETE CASCADE
+);
+CREATE TABLE sessions (
+  id SERIAL PRIMARY KEY,
+  "userId" INTEGER NOT NULL,
+  "sessionToken" VARCHAR(255) NOT NULL UNIQUE,
+  expires TIMESTAMPTZ NOT NULL,
+  CONSTRAINT fk_user FOREIGN KEY("userId") REFERENCES users(id) ON DELETE CASCADE
```
balazsorban44, line 28: Even when this is defined, the test fails with the following error:

> Could not find foreign keys between these entities. No relationship found between sessions and users

Here is the code that's failing:
```js
await client
        .from<AdapterSession & { user: AdapterUser }>("sessions")
        .select(`*, user: users(*)`)
        .eq("sessionToken", sessionToken)
        .single()
```

Anyone, ideas?

KushibikiMashu: How about this? 

```
-  .select(`*, user: users(*)`)
+  .select(`*, users:userId(*)`)
```

I hope this will help to solve that error.

balazsorban44: No, it did not seem to help, but thanks

p3lo: Just tested this:
```javascript
await supabase
      .from('sessions')
      .select(`*, user: userId(*)`)
      .eq('sessionToken', 'test_token')
      .single();
```

and I get result

```json
{
  "id": 1,
  "userId": 1,
  "sessionToken": "test_token",
  "expires": "2022-01-14T09:25:38+00:00",
  "user": {
    "id": 1,
    "name": "p3lo",
    "email": "p3lo@p3lo.com",
    "emailVerified": "2022-01-11T09:25:04+00:00",
    "image": "test"
  }
}
```

gabrielperales: @balazsorban44, is there any update on this? Does what @p3lo says fix the issue?

packages/supabase/src/index.ts
```diff
@@ -0,0 +1,176 @@
...
+        .from<Account>("accounts")
+        .select()
+        .eq("provider", provider)
+        .eq("providerAccountId", providerAccountId)
+        .single()
+
+      const userId = account.data?.userId
+      if (!userId) return null
+
+      return await this.getUser(userId)
```
gabrielperales, line 81: I've been testing this out. And seems like the reference to `this` is being lost. I have changed this line for `return adapter.getUser(userId)` and instead of returning the object, I have bonded it to a variable `adapter` and it works.


So I have something like:

```
const adapter = {
...
    async getUserByAccount({ provider, providerAccountId }) {
      const account = await client
        .from<Account>("accounts")
        .select()
        .eq("provider", provider)
        .eq("providerAccountId", providerAccountId)
        .single()

      const userId = account.data?.userId
      if (!userId) return null

      return adapter.getUser(userId)
},
...
}

return adapter
```
```

Comment (gabrielperales) [reply packages/supabase/src/index.ts:81]: this also seemed to work:

```
    async getUserByAccount({ provider, providerAccountId }) {
      const result = await client
        .from<Account>('accounts')
        .select('user: users(*)')
        .eq('provider', provider)
        .eq('providerAccountId', providerAccountId)
        .single()

      if (!result.data) return null

      return from<AdapterUser>(result.data.user)
    },
```

File: packages/supabase/src/index.ts [hunk 3]
```diff
@@ -0,0 +1,176 @@
...
+    return newObject
+  },
+}
+
+export function SupabaseAdapter(client: SupabaseClient): Adapter {
+  const { from, to } = formatter
+  return {
+    async createUser(data) {
+      const result = await client
+        .from<AdapterUser>("users")
```
gabrielperales, line 46: Why do we need to create a `users` table? shouldn't we be using the `auth.users` table? ...
So in this case we can use `supabase.auth.api.createUser` to create a user in that table.

More info: https://supabase.com/docs/reference/javascript/auth-api-createuser

packages/supabase/src/index.ts
gabrielperales: @balazsorban44 can you have a look ? thanks

gbarkhatov:
Any estimates on this?

balazsorban44:
I don't have the time to work on this, besides adapters have been migrated to the main monorepo. If someone is interested to take the code and opening a PR against https://github.com/nextauthjs/next-auth, please do!

Issue #192, opened 2021-08-26
Accept newest version of MongoDB driver `4.1.0`


italodeandra:
### Description 📓

The version 4 of the MongoDB driver is out.

### Error message ☕️

```
Error: optional dependency [mongodb] found but version [4.1.0] did not satisfy constraint [^3.5.9]
```

italodeandra:
The newest version of the driver looks really good. Before v4 the norm was to use Mongoose. But Mongoose is a ODM (similar to MikroORM and TypeORM), and I'm not sure to what point does NextAuth need that much features.

In my opinion we should be fine by using only the newest driver.

balazsorban44:
Question @italodeandra. Since mongoDB is not relational, I was wondering what would be a better structure for the data.

1. A single Users collection, where on each user we store an `accounts` and `sessions` array
2. or separate collections for Users, Accounts and Sessions, and connect them by a `userId` field.

Or either of these would be alright?

italodeandra:
I would suggest the second one to keep the idea of "relation" and avoid difference if someone tries to switch from one database provider to another while (somehow) keeping the data.

MongoDB you usually use the first option if the "sub-collection" is a primitive or something that would not grow (like user additional emails), but not for sessions since they can grow big. Even accounts I would suggest to also have a separate collection since it's not primitive, but that is just my preference.

balazsorban44:
@italodeandra I created an initial release here: https://www.npmjs.com/package/@next-auth/mongodb-adapter

Could you test it out? It requires `next-auth@4.0.0-next.24`. Some preliminary migration docs can be found here https://docs-nextauthjs.vercel.app/getting-started/upgrade-v4

I would be very grateful if you could have a look at the PR at #193 as well if you have the capacity. :pray: 

One thing, in particular, I struggle with is how to handle connections properly. Other solutions like Prisma or Fauna handle all that automatically, but in the case of MongoDB, it looks like we are forced to do it manually. Is this correct?

If I also go with option 2 from https://github.com/nextauthjs/adapters/issues/192#issuecomment-907668413 (which I did in the above release), then I'll have to find out how to do transactions to delete Sessions and Accounts when a user is deleted.
Because this is not necessarily atomic:
https://github.com/nextauthjs/adapters/blob/cf0e69f2d4fab984bff285830e35cd00ae46b740/packages/mongodb/src/index.ts#L87-L91

[...]

PR #193, opened 2021-08-28
feat(mongodb): add standalone mongoDB adapter

balazsorban44:
This will partially replace the TypeORM adapter that has been trying to be too many things in one. There is an SQL adapter counterpart at #194.

Related issues: #178, #192 

~One issue still not solved is how to handle connections nicely.~

Decided to follow the example from the Next.js repo, as probably most people have that kind of setup already: https://github.com/vercel/next.js/tree/canary/examples/with-mongodb


packages/mongodb/src/index.ts
```diff
@@ -0,0 +1,148 @@
...
+    async updateUser(data) {
+      const { value: user } = await Users.findOneAndUpdate(
+        { _id: _id(data.id) },
+        { $set: data }
+      )
+      return from<AdapterUser>(user!)
+    },
+    async deleteUser(id) {
+      await Promise.all([
+        m.collection(collections.Accounts).deleteMany({ userId: id }),
```
italodeandra, line 87: The relations doesn't have the string converter. Does that mean that the relations are string primitives?

balazsorban44: Could you explain what you mean here exactly? 😊

italodeandra: The **ObjectId** consists of a 12-byte value (4-byte is a creation timestamp, 5-byte is a random value, 3-byte is a initially random incrementing counter), but its hexadecimal representation (the string), depending on the charset you're using, can consist of around 1-byte per character while having 24 characters which means double the size.

The **ObjectId** not only have a better performance for indexing, but also takes less space if you're not indexing. For me converting is a must because of those two reasons.

balazsorban44: Fixed!

italodeandra: There's one thing I forgot to mention that might be a problem, I'm not really sure how are you handling this on other cases.

The **ObjectId** is not serializable by NextJS. So if you try to pass it on the `getServerSideProps`, `getInitialProps`, and `getStaticProps` it will throw an error.

Since I already use `mongodb` on my application, I handle that by transforming my data before dehydrating it.

On the TypeORM providers for example, do you return a lean version of the objects or do you return a TypeORM model? If you just return the TypeORM model, it might be a good idea to transform the **ObjectId** to hexadecimal before returning or add a note on the documentation.

---

Some references:

https://github.com/blitz-js/superjson/issues/69
https://github.com/blitz-js/superjson#using-with-nextjs
https://github.com/blitz-js/superjson/pull/71

Disclaimer: I'm not suggesting to use `superjson` or anything, it's just the closest thing I found to help explain the issue.

italodeandra: Oh well, it looks you're doing that already. Nice!

balazsorban44: Yes, before sending it to the core, I transform the ids to hex strings.

balazsorban44:
@italodeandra if you want to give it a try with the latest fixes, you can use the version `@next-auth/mongodb-adapter@0.0.2-pr.193-4f0ef7e6.282`

I also removed the requirement to pass `ObjectId` as an option to the adapter. I figured if we have `mongodb` as a peer dependency, it will be resolved correctly, even if I import `ObjectId` in the adapter. 😄

italodeandra:
> @italodeandra if you want to give it a try with the latest fixes, you can use the version `@next-auth/mongodb-adapter@0.0.2-pr.193-4f0ef7e6.282`

I'll do it now.

> I also removed the requirement to pass `ObjectId` as an option to the adapter. I figured if we have `mongodb` as a peer dependency, it will be resolved correctly, even if I import `ObjectId` in the adapter. 😄

Nice find! I have library and did that already by reflex (everything that is not specific to that library I make peer dependency) so I didn't even think of it. The first time I reviewed I thought `Hmm.. Okay, makes sense, he probably don't want to import mongodb just because of the ObjectId`. TIL.

italodeandra (approved):
Code looks all good.

Tested on real application and working fine.

italodeandra:
> I figured if we have `mongodb` as a peer dependency, it will be resolved correctly, even if I import `ObjectId` in the adapter.

Not sure if it changes anything for you but `mongodb` uses **ObjectId** from `bson` (as a peer dependency). I just remembered that and I thought that you should know that if you don't already.

balazsorban44:
`mongodb` still has to be a peer dependency because setting up the connection has to be done by the user, so I don't think it changes anything, but thanks!

ndom91 (approved):
LGTM :+1:

PR #185, opened 2021-08-18
feat(typeorm): use new Adapter API

balazsorban44:
See #171, #182, #183

> NOTE: from v4, mongoDB will have its own adapter (see #193), and TypeORM will only support SQL flavoured databases

TODO: Docs

PR #194, opened 2021-08-29
feat(knex): add standalone SQL adapter 

ndom91:
Smaller more well-maintained replacement for the TypeORM adapter for v4, rewritten to match the simplified adapter API.

The easiest ORM we found was `knex`, so this adapter will use that under the hood.

A separate MongoDB adapter can be found at #193


balazsorban44:
Closing in favor of https://github.com/nextauthjs/adapters/pull/185 after discussion.

PR #206, opened 2021-09-05
feat(dgaph): add Dgraph adapter

arnaudDerbey:
- [x] Documentation
- [x] Tests
- [ ] Ready to be merged

This where I'm at about the dgraph adapter for v4. I still need to test it IRL and clean it up.


balazsorban44:
this repository uses `lerna`, which itself requires `yarn`. could you remove the `package-lock.json` file and run `yarn` to install dependencies?

arnaudDerbey:
will do !

balazsorban44:
There seems to be another `package-lock.json` in the root folder as well

staticle:
Hi I had a sanity check on the code and it looks cool can't wait to try it out please merge

balazsorban44:
We will need more than sanity checking before merging, but it is on the agenda! :wink:

If you want to use it right now, you could simply copy the code in this PR!

balazsorban44:
@arnaudDerbey running the tests locally, I get some errors, would you be able to fix those before I review your code?

bmartel:
Is there any movement on this? It appears the tests now pass for this adapter. 
image

balazsorban44:
I'll have a look, sorry for the delay. I was getting the core repo ready for stable v4 in the past couple of weeks, but most of the important PRs have been merged, I'll focus on adapters and the docs in the coming weeks!

balazsorban44:
@bmartel if interested in helping:

> An accompanying documentation page should also be added in https://github.com/nextauthjs/docs/tree/main/docs/adapters

> See nextauthjs/docs#37 for inspiration

packages/dgraph/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "private": false,
+  "publishConfig": {
+    "access": "public"
+  },
+  "scripts": {
+    "build": "tsc",
+    "test": "jest"
+  },
+  "peerDependencies": {
+    "next-auth": ">4 || 4.0.0-beta.1 - 4.0.0-beta.x",
```
balazsorban44, line 33: Corrected as per #274

balazsorban44:
@arnaudDerbey could you add dgraph to the labeler in https://github.com/nextauthjs/adapters/blob/1695fa3bb4b6a2122506db8f3f4e952622ebc319/.github/labeler.yml

bmartel:
> @bmartel if interested in helping:
> 
> > An accompanying documentation page should also be added in https://github.com/nextauthjs/docs/tree/main/docs/adapters
> 
> > See [nextauthjs/docs#37](https://github.com/nextauthjs/docs/pull/37) for inspiration
> 
> Other than that, indeed tests seem to pass now, so as an initial release this should be good to go! When there is a PR to the docs, I'll merge this.
> 
> I appreciate it if @arnaudDebrey and @bmartel will be around in ~2 weeks when v4 goes stable in case any problem should arise.

I PR'd the docs additions. I realized that the imports for the adapter were mentioned in the package as `import { DgraphAdapter, DgraphClient } from '@next-auth/dgraph-dapter/dgraphAdapter'`. I corrected this in the documentation, however I just want to double check this is properly exported in the package code @arnaudDerbey.

bmartel:
I will be around in the coming weeks @balazsorban44 and can assist in getting this across the finish line if needed.

balazsorban44:
Some things that came up while reviewing this with @ndom91 today:

- Why aren't you using https://dgraph.io/docs/clients/javascript/http/? It seems that this could make things much easier
- The whole `dGraphClient.ts` file seems to be unnecessary, you basically re-implement the whole Adapter. I recommend removing that file and just writing the logic directly in the main file. If you use the above-mentioned library, some abstractions could be moved into another file maybe, but really not much
- The fragments here seem to be corresponding to models/schemas. they should be written in a way that is easy to extend if the user needs to, maybe even generate those queries dynamically, from the given data objects.

All in all, it is a good start, but there is potential for improvement.

I think the most important thing before merging is to decide on the adapter method signature (should probably be using the dgraph JS client and _that_ should be passed as an argument), and that all tests pass. After that, we could do a cleanup in a separate PR

bmartel:
Hey @balazsorban44, I agree with most of what you mentioned as feedback for @arnaudDerbey with regards to reuse an existing http client, and write this in a way that allows for extension. The part that I don't agree with is the last line about the dgraph client being used. There are 2 ways to interact with dgraph, 1 through graphql, and the other through dgraph's syntax (what they have historically called graphql+-). The aforementioned is not compatible with graphql and is the sole form of which the official dgraph js clients operate. I do not believe these adapters would work in this case. Also I don't believe it is a great idea to opt for the usage of graphql+- to enable use of the official dgraph clients, as it requires the end user to understand a lesser used format which is normally for advanced cases only.

I believe the correct way forward for dgraph would be to reuse an existing http client, write the queries/mutations in a enduser configurable way so overrides or augmentations may take place.

arnaudDerbey:
hi all ! sorry, lots of work during the past few months. :) I will check where we're at this week end.

arnaudDerbey:
@bmartel is right the dgraph js client is for dql syntax and this adapter was built for graphql users. Using dql is the job of another adapter.

You're right @balazsorban44 the whole dGraphClient.ts is pure overhead and add useless complexity. Will remove it. 

Having userdefined fragment is a very good idea !

balazsorban44 (changes requested):
Please have a look at other adapters like

https://github.com/nextauthjs/adapters/tree/next/packages/neo4j

https://github.com/nextauthjs/adapters/tree/next/packages/fauna

To see how we format data back and forth. Especially thinking of how you manually coded `emailVerifierd`.

Remove any `updatedAt` and `createdAt` from the code. These aren't necessary, and our official adapter should be as simple as possible. Whenever possible, don't use try catch, let the errors bubble up and be handled by the core.

I also suggest removing the `getAccount` and  `getVerificationRequest` (name is wrong, should be `getverificationToken` anyway). The `client` you are creating should have no involvement in the actual adapter.

`clean` should not be part of the actual adapter/client code either.

You can just run that logic directly in the test file.


Tests are also failing for me locally.

Feel free to fix those first, and I will be happy to merge it in, and do the cleanup myself afterward if that's fine by you.

arnaudDerbey:
> Especially thinking of how you manually coded `emailVerifierd`.

Do you refer to things like this ? ` return { ...user, emailVerified: new Date(user.emailVerified) }` this is because Dgraph return a string and not a date object.

balazsorban44:
Yes, I do. I understand that. But rather than formatting only that field, you should wrap the result with something like `format.from` that maps the values, regardless of their names.

For example: https://github.com/nextauthjs/adapters/blob/ef1bff04788505105f119c1f1b7457dea3338566/packages/neo4j/src/utils.ts#L12-L41

arnaudDerbey:
Isn't it a bit too much just to parse a string to a date ?

balazsorban44:
Yeah, but it is robust enough in case there are other formats to parse. The main thing is that you don't need the name of the fields that should be formatted.

lveillard:
Hello guys! Just to tell you, i'm actively using this package, so i will be available to test things for you if needed before you launch the beta :). I've copied the last version into my project, added the custom decode/encode logic and it works like a charm. Documentation was clear and I manage to do everything really fast. 

I'm a bit more scared of the secured version, but will tell you how it goes also

First quick feedback is that as i'm using the JWT version, i don't need the session type, maybe we could add a comment to tell people the optional stuff of the schema depending on the database / jwt session? (unless i'm missing somthing)

Thanks for the great work both the next-auth team and Arnaud btw

balazsorban44 (changes requested):
verification request is still a misleading name. It should be verification token.

The name stuck because, in previous versions, the adapter had the responsibility to send the token to the e-mail address, but this is not the case anymore. The adapter is a simple data broker between the core and the database.

balazsorban44 (approved):
I am going to approve/merge this. I would like to release v4 on Monday, and since this is passing the tests, I feel confident to merge, but follow the upcoming PR for my suggested changes.

arnaudDerbey:
Ok I will. 👍

balazsorban44:
Quick question. why is `getuserByEmail` returning an array? We don't expect two users to have the same e-mail.

PR #224, opened 2021-09-14
feat(typeorm): Refactor getManager

### LINKED ISSUE #222: [V4] TypeORM-legacy adapter connection not being set correctly
I believe the next version of [typeorm-legacy-adapter ](https://github.com/nextauthjs/adapters/tree/next/packages/typeorm-legacy) has a broken connection manager. 

Currently, it looks like the connection will never be able to hold the `default` connection, as it always defaults to trying to create a new connection. Not exactly sure why this happens. This results in the `A connection with the name "default" already exists` error

I solved this by writing a utility function `ensureConnection` and also add the "has entities changed" check.

```typescript
function entitiesChanged(prevEntities: any, newEntities: any[]): boolean {
  if (prevEntities?.length !== newEntities?.length) return true;

  for (let i = 0; i < prevEntities?.length; i++) {
    if (prevEntities[i] !== newEntities[i]) return true;
  }

  return false;
}

export async function updateConnectionEntities(connection: Connection, entities: any[]) {
  if (!entitiesChanged(connection.options.entities, entities)) return;

  // @ts-ignore
  connection.options.entities = entities;

  // @ts-ignore
  connection.buildMetadatas();

  if (connection.options.synchronize) {
    await connection.synchronize()
  }
}

export async function ensureConnection(_connection: Connection, config: ConnectionOptions & { entities: any[]}) {
  const connectionManager = getConnectionManager();

  if (connectionManager.has('default')) {
    _connection = connectionManager.get();

    _connection.isConnected || (await _connection.connect());

    if (process.env.NODE_ENV !== 'production') {
      await updateConnectionEntities(_connection, config.entities);
    }
  } else {
    _connection = await connectionManager.create(config).connect();
  }

  return _connection.manager;
}

```

**getManager** now looks like this
```typescript
...
export async function getManager(options: {
  connection: string | ConnectionOptions;
  entities: Entities;
}): Promise<EntityManager> {
  const { connection, entities } = options;

  const config = {
    ...parseConnectionConfig(connection),
    entities: Object.values(entities)
  };

  return await ensureConnection(_connection, config);
}
...
```

Fronix:
## Reasoning 💡

The current implementation of `getAdapter` doesn't let you use adapter functions outside the next-auth package. Example would be: Doing a database lookup in the `signIn` callback, it will fail due to a connection already existing with the name `default`.

This allows you to use adapter functions outside next-auth package without that error.

It also adds back the old functionality of V3 typeorm adapter to automatically sync entity changes by default when you change entity files and also updates typeorm metadata. You can disable this by setting `synchronize: false`

## Checklist 🧢

- [x] Documentation
- [x] Tests
- [ ] Ready to be merged

## Affected issues 🎟

Fixes #222

### DIFF
--- a/basic-tests.ts
+++ b/basic-tests.ts
@@ -1,5 +1,5 @@
 import { Adapter } from "next-auth/adapters"
-import { createHash, randomUUID as cryptoRandomUUID, randomBytes } from "crypto"
+import { createHash, randomUUID } from "crypto"
 
 export interface TestOptions {
   adapter: Adapter
@@ -308,10 +308,7 @@ export function hashToken(token: string) {
   return createHash("sha256").update(`${token}anything`).digest("hex")
 }
 
-function randomUUID() {
-  // Use `randomUUID` if available. (Node 15.6++)
-  return cryptoRandomUUID?.() ?? randomBytes(16).toString("hex")
-}
+export { randomUUID }
 
 export const ONE_WEEK_FROM_NOW = new Date(Date.now() + 1000 * 60 * 60 * 24 * 7)
 export const FIFTEEN_MINUTES_FROM_NOW = new Date(Date.now() + 15 * 60 * 1000)
--- a/basic-tests.ts
+++ b/basic-tests.ts
@@ -1,5 +1,5 @@
 import { Adapter } from "next-auth/adapters"
-import { createHash, randomUUID } from "crypto"
+import { createHash, randomUUID as cryptoRandomUUID, randomBytes } from "crypto"
 
 export interface TestOptions {
   adapter: Adapter
@@ -308,7 +308,10 @@ export function hashToken(token: string) {
   return createHash("sha256").update(`${token}anything`).digest("hex")
 }
 
-export { randomUUID }
+function randomUUID() {
+  // Use `randomUUID` if available. (Node 15.6++)
+  return cryptoRandomUUID?.() ?? randomBytes(16).toString("hex")
+}
 
 export const ONE_WEEK_FROM_NOW = new Date(Date.now() + 1000 * 60 * 60 * 24 * 7)
 export const FIFTEEN_MINUTES_FROM_NOW = new Date(Date.now() + 15 * 60 * 1000)
--- a/packages/typeorm-legacy/src/utils.ts
+++ b/packages/typeorm-legacy/src/utils.ts
@@ -101,7 +101,10 @@ export async function updateConnectionEntities(
   connection.buildMetadatas()
 
   if (connection.options.synchronize !== false) {
-    console.warn("[next-auth] ADAPTER_TYPEORM_UPDATING_ENTITIES")
+    console.warn(
+      "[next-auth][warn][adapter_typeorm_updating_entities]",
+      "\nhttps://next-auth.js.org/warnings#adapter_typeorm_updating_entities"
+    )
     await connection.synchronize()
   }
 }
--- a/packages/typeorm-legacy/src/utils.ts
+++ b/packages/typeorm-legacy/src/utils.ts
@@ -1,4 +1,4 @@
-import { ConnectionOptions } from "typeorm"
+import { Connection, ConnectionOptions } from "typeorm"
 import * as defaultEntities from "./entities"
 
 /** Ensure configOrString is normalized to an object. */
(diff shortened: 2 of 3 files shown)


Fronix:
`mysql `and `postgresql `tests work fine, but `sqlite` test break because `name` is set. If name is set in either test, they do break. Not sure why.

If no `config.name` is set it should be `default` but it's like name never gets passed if set in config.

balazsorban44:
All tests should pass before merging though. :grimacing: there should be a way to get around it.

basic-tests.ts
```diff
@@ -308,7 +308,10 @@ export function hashToken(token: string) {
   return createHash("sha256").update(`${token}anything`).digest("hex")
 }
 
-export { randomUUID }
+function randomUUID() {
+  // Use `randomUUID` if available. (Node 15.6++)
+  return cryptoRandomUUID?.() ?? randomBytes(16).toString("hex")
+}
```
balazsorban44, line 314: Could you please revert these changes? You should be able to get around this if you install Node 16. We don't wish to maintain code for multiple Node versions for the sake of testing.

packages/typeorm-legacy/src/utils.ts
```diff
@@ -74,3 +74,35 @@ export function parseConnectionConfig(
...
+  // @ts-expect-error
+  connection.options.entities = entities
+
+  console.warn("[next-auth] ADAPTER_TYPEORM_REBUILDING_METADATA")
+  // @ts-expect-error
+  connection.buildMetadatas()
+
+  if (connection.options.synchronize) {
+    console.warn("[next-auth] ADAPTER_TYPEORM_UPDATING_ENTITIES")
+    await connection.synchronize()
```
balazsorban44, line 106: I thought we wanted to do this regardless of user config, if `process.env.NODE_ENV !== "production"`. I was actually fine by it. Also, one `console.warn` is probably enough.

balazsorban44 (changes requested):

Fronix:
> All tests should pass before merging though. 😬 there should be a way to get around it.

Yeah, I'll have to take a look tomorrow. Didn't have enough time to dig into why `name` isn't being set from the config correctly. I'll take a closer look tomorrow

packages/typeorm-legacy/src/utils.ts
Fronix: This would effectively disable the `synchronize` option, I feel like that could be bad? Removes the option to not `synchronize` if you need to test something.

That's why I put it here https://github.com/nextauthjs/adapters/blob/faa2c76f28b99b6f5baecac5f95dde56eb9a7af7/packages/typeorm-legacy/src/index.ts#L40 since `updateConnectionEntities` is only useful when not in production.

Yeah, one warning is enough, I agree.

balazsorban44: OK, I see. What if we did
`if (connection.options.synchronize !== false)`?

So basically sync by default? The user could still disable in not prod if needed?

Fronix:
@balazsorban44 how do you want me to handle the documentation part?

balazsorban44:
The only relevant change in the docs is the added warning, I think. You can find the source here:

https://github.com/nextauthjs/docs/blob/main/docs/warnings.md

Maybe adding a new Adapter section after Client and Server, and just describe the warning there

balazsorban44 (approved):
Thank you so much for this PR! I ran tests locally, and it worked nicely! :pray: :green_heart:

packages/typeorm-legacy/src/utils.ts
```diff
@@ -74,3 +74,34 @@ export function parseConnectionConfig(
...
+  if (!entitiesChanged(connection.options.entities, entities)) return
+
+  // @ts-expect-error
+  connection.options.entities = entities
+
+  // @ts-expect-error
+  connection.buildMetadatas()
+
+  if (connection.options.synchronize !== false) {
+    console.warn("[next-auth] ADAPTER_TYPEORM_UPDATING_ENTITIES")
```
balazsorban44, line 104:
```suggestion
    console.warn(
      "[next-auth][warn][adapter_typeorm_updating_entities]",
      "\nhttps://next-auth.js.org/warnings#adapter_typeorm_updating_entities"
    )
```
Just a small nitpick, to make this easier for people, let's add a link!

balazsorban44 (changes requested):
I added a ver small suggestion, could you address it, and then this is good to go!

packages/typeorm-legacy/src/utils.ts
Fronix: I had just thought of that and was going to add it when I had the time. Great 👍

balazsorban44 (approved):

balazsorban44:
@Fronix this is now released in `@next-auth/typeorm-legacy-adapter@0.1.2-next.11`!

PR #234, opened 2021-09-18
fix(prisma): bump Prisma to v3

balazsorban44:
This PR bumps `prisma` to `v3`. In our case, it did not introduce any breaking changes.

We also added tests for custom user properties and changed the Prisma test run in a way so we could run `jest` in watch mode, for faster iteration.

We achieve it by cleaning up the database between each test run, instead of doing it at the beginning of the `test` script only.

This should be added to ALL other adapters, so we gain confidence in supporting custom models.

In addition, @ndom91 @lluia and I discussed that `basic-tests` should be refactored so each test is isolated and doesn't depend on previous tests to succeed, which can be its own separate PR and would result in easier debugging while working on an adapter. #235


ndom91 (approved):
LGTM 👍

PR #73, opened 2021-04-28
feat(sequelize): initial adapter

EXIT-ALAMERY:
Added the adapter for Sequelize ORM.

PR #237, opened 2021-09-19
feat(sequelize): Add Sequelize adapter

luke-j:
## Reasoning 💡

This follows on from https://github.com/nextauthjs/adapters/pull/73 (although doesn't share much code) and creates an adapter for [Sequelize](https://sequelize.org). Appreciate any feedback!

## Checklist 🧢

- [x] Documentation
- [x] Tests
- [ ] Ready to be merged

I still have to test this more thoroughly locally, but would appreciate any feedback in the meantime.


balazsorban44:
Looks promising, but running the tests from https://github.com/nextauthjs/adapters/blob/next/basic-tests.ts would be crucial for this to be merged. Keep up the good work, I'll review when you manage to run all tests successfully :wink: Feel free to tell me if you are stuck somewhere.

luke-j:
No problems @balazsorban44 

>running the tests from https://github.com/nextauthjs/adapters/blob/next/basic-tests.ts would be crucial for this to be merged.

Could you clarify? I've added these tests and they're passing for me locally, looks like the GitHub action requires maintainer approval to run.

balazsorban44:
My bad, I skimmed through https://github.com/nextauthjs/adapters/blob/748d7dc8fb4d8188d81b5e171a21943369ada642/packages/sequelize/tests/index.test.ts :sweat_smile: You are right about GitHub action requiring a maintainer. I'll test this locally then!

packages/sequelize/src/index.ts
```diff
@@ -0,0 +1,156 @@
...
+  const Session =
+    (client.models.Session as ModelCtor<any>) ||
+    client.define<SessionInstance>("session", sessionSchema, tableOptions)
+  const VerificationToken =
+    (client.models.VerificationToken as ModelCtor<any>) ||
+    client.define<VerificationTokenInstance>(
+      "verificationToken",
+      verificationTokenSchema,
+      tableOptions
+    )
```
balazsorban44, line 50:
```suggestion
import type { Sequelize, Model, ModelCtor, ModelOptions } from "sequelize"
import * as defaultModels from "./schema"

export { defaultModels as models }

// @see https://sequelize.org/master/manual/typescript.html
interface AccountInstance
  extends Model<ApadterAccount, Partial<ApadterAccount>>,
    ApadterAccount {}
interface UserInstance
  extends Model<AdapterUser, Partial<AdapterUser>>,
    AdapterUser {}
interface SessionInstance
  extends Model<AdapterSession, Partial<AdapterSession>>,
    AdapterSession {}
interface VerificationTokenInstance
  extends Model<VerificationToken, Partial<VerificationToken>>,
    VerificationToken {}

interface SequelizeAdapterOptions {
  models?: Partial<{
    User: ModelCtor<UserInstance>
    Account: ModelCtor<AccountInstance>
    Session: ModelCtor<SessionInstance>
    VerificationToken: ModelCtor<VerificationTokenInstance>
  }>
}

export default function SequelizeAdapter(
  client: Sequelize,
  options?: SequelizeAdapterOptions
): Adapter {
  const { models } = options ?? {}

  const defaultModelOptions: ModelOptions = {
    underscored: true,
    timestamps: false,
  }

  const { User, Account, Session, VerificationToken } = {
    User:
      models?.User ??
      client.define<UserInstance>(
        "user",
        defaultModels.User,
        defaultModelOptions
      ),
    Account:
      models?.Account ??
      client.define<AccountInstance>(
        "account",
        defaultModels.Account,
        defaultModelOptions
      ),
    Session:
      models?.Session ??
      client.define<SessionInstance>(
        "session",
        defaultModels.Session,
        defaultModelOptions
      ),
    VerificationToken:
      models?.VerificationToken ??
      client.define<VerificationTokenInstance>(
        "verificationToken",
        defaultModels.VerificationToken,
        defaultModelOptions
      ),
  }
```

I would refactor this section in a way that makes it easier to configure by the user.

With my suggestions, the user would override `User` like this:

```js
SequelizeAdapter(client, {
  schemas: {
    User: client.define("user", MyUserModel)
  }
})
```

What do you think?

Also for consistent naming, I think we should just stick to models/model instead of schemas, as the TS types suggest that anyway. (could you rename the `schema.ts` file to `models.ts` as well?)

packages/sequelize/src/index.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  Account.belongsTo(User, { onDelete: "cascade" })
+  Session.belongsTo(User, { onDelete: "cascade" })
```
balazsorban44, line 53: Is this necessarily done runtime? I am not familiar with Sequelize, but usually, you would do some kind of seeding/setup step to define the database schema, right? Does Sequelize have anything similar to TypeORM's `syncrhonize: true` (https://github.com/typeorm/typeorm/blob/master/docs/faq.md#how-do-i-update-a-database-schema)?

We have to be careful not to update things in production to prevent data loss. In development though, auto-syncing model changes might be useful. #224 how we handled this for the TypeORM adapter recently.

packages/sequelize/src/schema.ts
```diff
@@ -0,0 +1,39 @@
...
+  expires: { type: DataTypes.DATE, allowNull: false },
+  sessionToken: { type: DataTypes.STRING, unique: true, allowNull: false },
+  userId: { type: DataTypes.INTEGER },
+}
+
+export const verificationTokenSchema = {
+  token: { type: DataTypes.STRING, primaryKey: true },
+  identifier: { type: DataTypes.STRING, allowNull: false },
+  expires: { type: DataTypes.DATE, allowNull: false },
+}
```
balazsorban44, line 39:
```suggestion
export const User = {
  id: { type: DataTypes.INTEGER, primaryKey: true, autoIncrement: true },
  name: { type: DataTypes.STRING },
  email: { type: DataTypes.STRING, unique: true },
  emailVerified: { type: DataTypes.DATE },
  image: { type: DataTypes.STRING },
}

export const Account = {
  id: { type: DataTypes.INTEGER, primaryKey: true, autoIncrement: true },
  type: { type: DataTypes.STRING, allowNull: false },
  provider: { type: DataTypes.STRING, allowNull: false },
  providerAccountId: { type: DataTypes.STRING, allowNull: false },
  refresh_token: { type: DataTypes.STRING },
  access_token: { type: DataTypes.STRING },
  expires_at: { type: DataTypes.INTEGER },
  token_type: { type: DataTypes.STRING },
  scope: { type: DataTypes.STRING },
  id_token: { type: DataTypes.STRING },
  oauth_token_secret: { type: DataTypes.STRING },
  oauth_token: { type: DataTypes.STRING },
  session_state: { type: DataTypes.STRING },
  userId: { type: DataTypes.INTEGER },
}

export const Session = {
  id: { type: DataTypes.INTEGER, primaryKey: true, autoIncrement: true },
  expires: { type: DataTypes.DATE, allowNull: false },
  sessionToken: { type: DataTypes.STRING, unique: true, allowNull: false },
  userId: { type: DataTypes.INTEGER },
}

export const VerificationToken = {
  token: { type: DataTypes.STRING, primaryKey: true },
  identifier: { type: DataTypes.STRING, allowNull: false },
  expires: { type: DataTypes.DATE, allowNull: false },
}
```

Let's rename this file to `models.ts`.

Could we use `uuid` or similar for `id` on all of these models? We prefer that on other adapters as well, for consistency, and I assume it's a tiny bit more secure if ID's are not sequential/easier to guess anyway.

packages/sequelize/tests/index.test.ts
```diff
@@ -0,0 +1,45 @@
...
+  adapter: SequelizeAdapter(sequelize),
+  db: {
+    connect: async () => {
+      return await sequelize.sync({ force: true })
+    },
+    verificationToken: async ({ identifier, token }) => {
+      const verificationToken =
+        await sequelize.models.verificationToken.findOne({
+          where: { identifier, token },
+        })
```
balazsorban44, line 21:
```suggestion
    verificationToken: async (where) => {
      const verificationToken =
        await sequelize.models.verificationToken.findOne({ where })
```

packages/sequelize/tests/index.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    },
+    user: async (id) => {
+      const user = await sequelize.models.user.findByPk(id)
+
+      return user?.get({ plain: true }) || null
+    },
+    account: async ({ provider, providerAccountId }) => {
+      const account = await sequelize.models.account.findOne({
+        where: { provider, providerAccountId },
+      })
```
balazsorban44, line 33:
```suggestion
    account: async (where) => {
      const account = await sequelize.models.account.findOne({ where })
```

packages/sequelize/src/index.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  return {
+    async createUser(user) {
+      return await User.create(user)
+    },
+    async getUser(id) {
+      const userInstance = await User.findByPk(id)
+
+      return userInstance?.get({ plain: true }) ?? null
```
balazsorban44, line 62: I love `plain: true` :heart_eyes:! This reminds me of the Prisma Adapter, which is by far the easiest to understand IMO, since it can return plain JavaScript objects, instead of convoluted custom formats for dates and stuff.

All ORMs should have an option like this! :pray:

packages/sequelize/src/index.ts
```diff
@@ -0,0 +1,156 @@
...
+    },
+    async updateUser(user) {
+      await User.update(user, { where: { id: user.id } })
+      const userInstance = await User.findByPk(user.id)
+
+      if (!userInstance) {
+        throw new Error("User does not exist")
+      }
+
+      return userInstance
```
balazsorban44, line 92:
```suggestion
      return userInstance!
```

You could assume this never happens because we won't try updating a user from the core that does not exist.

(you will need to add `/* eslint-disable @typescript-eslint/no-non-null-assertion */` to the top of the file to be able to use the `!` TS non-null assertion)

balazsorban44 (changes requested):
Very nice work! I only had a few suggestions! It looks to me that Sequelize is very nice to work with. 👌

I could see it as a replacement in our suggestions for the TypeORM adapter IMO.

packages/sequelize/README.md
```diff
@@ -0,0 +1,63 @@
...
+// https://next-auth.js.org/configuration/options
+export default NextAuth({
+  // https://next-auth.js.org/configuration/providers
+  providers: [
+    Providers.Google({
+      clientId: process.env.GOOGLE_ID,
+      clientSecret: process.env.GOOGLE_SECRET,
+    }),
+  ],
+  adapter: SequelizeAdapter(sequelize)
```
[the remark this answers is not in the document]
luke-j: Yep happy to submit a docs PR once this PR is finalised 👍

packages/sequelize/src/index.ts
luke-j: Sequelize has the [`sequelize.sync()`](https://sequelize.org/master/manual/model-basics.html#model-synchronization) function similar to TypeORM's synchronize option. Defining these associations at runtime is recommended and doesn't produce any SQL unless `sync()` is called.

It would be best practice for the user to define the tables as migrations, but sequelize will automatically create the necessary tables/FKs/indexes with `sync` like so:


```
const sequelize = new Sequelize('sqlite:memory:')
const adapter = SequelizeAdapter(sequelize)

sequelize.sync()

export default NextAuth({
  ...
  adapter
  ...
})
```

I've added a section in the README about how to create tables like this with a mention that best practice is still migrations in e0b1dae.

packages/sequelize/src/schema.ts
luke-j: 👍  I've used uuids instead in 30363a1

packages/sequelize/src/index.ts
luke-j: Yep I think that's a fair change, I've refactored that in f6b9518 I've also added docs on how to do this in cd44414

packages/sequelize/src/index.ts
luke-j: Agree! Having an instance returned can be useful but the option for a plain object is always great.

packages/sequelize/src/index.ts
luke-j: Nice, done in 3b2da67

luke-j:
Thanks for the excellent feedback @balazsorban44.

>I could see it as a replacement in our suggestions for the TypeORM adapter IMO.

Sequelize is certainly the most mature ORM available at the moment, I think. It's not as new and shiny as TypeORM but it's a solid choice IMO. I'd be happy to help out with any refactor 👍

packages/sequelize/src/index.ts
balazsorban44: I think we should default to synchronizing if we detect `process.env.NODE_ENV !== "production"`, but make it so that users can opt-out even then, with a config option like `synchronize: false`.

We could reuse this warning by some rewording: https://next-auth.js.org/warnings#adapter_typeorm_updating_entities

I would add `client.sync()` in an if statement with the warning	similar to this [`c2559b7` (#224)](https://github.com/nextauthjs/adapters/pull/224/commits/c2559b7557e7786db8dc96d7f545a9c80c79af18) in the body of the Adapter.

Otherwise, we should create explain how to migrate manually: https://sequelize.org/master/manual/migrations.html

luke-j: @balazsorban44 I'm thinking about how we might achieve this. I remembered that `sync` actually returns a promise, in which case we'd have to make the adapter `async` which would mess with the user-facing API.

eg.
```ts
if (process.env.NODE_ENV !== 'production' && synchronize) {
  await Promise.all([
    User.sync(),
    Account.sync(),
    Session.sync(),
    VerificationToken.sync()
  ])
}
```

I'm wondering if including `sync`, which is potentially dangerous or unexpected from the user's perspective, is worth it in this regard? There's not a huge advantage over just leaving this to user land code, right?

balazsorban44: I see. Ultimately, this is why we had to get a connection manager in each adapter method with TypeORM, because getting a manager was an async operation. :sweat_smile: 

https://github.com/nextauthjs/adapters/blob/next/packages/typeorm-legacy/src/index.ts

I don't want the adapter to be async as that would complicate the initialization.

I see there is an all-in-one `.sync()` method as well.

https://sequelize.org/master/class/lib/sequelize.js~Sequelize.html#instance-method-sync

My idea is that we should create a wrapper for it, and add a `sync: boolean | SyncOptions` to the adapter options, and do something similar to TypeORM, and call it in each method first.

I feel like we should stretch here, as the DX would be better. Especially useful if you would need to debug, I think.

I am thinking of something like this:
```ts
function SequelizeAdapter(client, options) {
  // Make sure we only sync once per invocation
  let _synced = false
  const sync = async () => {
    if (
      process.env.NODE_ENV === "production" ||
      _synced ||
      options.sync === false
    )
      return

    const syncOptions =
      typeof options.sync === "object"
        ? options.sync
        : {
            // ... some default options, if we need it
          }
    await client.sync(syncOptions)
    _synced = true
  }

  return {
    async createUser() {
      await sync()
      //...
    },
  }
}
```

packages/sequelize/README.md
```diff
@@ -0,0 +1,99 @@
...
+const adapter = SequelizeAdapter(sequelize)
+
+// Calling sync() is not recommended in production
+sequelize.sync()
+
+export default NextAuth({
+  ...
+  adapter
+  ...
+})
```
balazsorban44, line 67:
```suggestion

export default async function auth(req, res) {
    await sequelize.sync()
	return await NextAuth(req, res, {
	  ...
      adapter
      ...
	})
}
```

As you pointed out, `.sync()` returns a Promise, so I guess this wouldn't work anyway. I leave a suggestion just to show you the alternative, but I don't think we should actually do it that way. See my comment here: https://github.com/nextauthjs/adapters/pull/237/files#r712815434

packages/sequelize/src/index.ts
luke-j: Yeah I think if we want to do this this is probably the best option. I've done this in 48961d1. Unfortunately it's not the prettiest solution.

If this is occurring in multiple adapters I think it's reasonable to consider an optional `connect` or `init` function returned from an adapter which is then called by the main next-auth package prior to calling the other functions.

balazsorban44: Most adapters seem to connect/disconnect automatically, actually. In my experience, it is only TypeORM that needs manual connection handling so far. :confused:

balazsorban44 (changes requested):
Thanks, just a few more suggestions.

rageshoo:
@luke-j Hi Luke, thanks for working on this! Do you have any updates?

luke-j:
@balazsorban44 I've defaulted `synchronize` to true in a03a37c & 564dc81. Let me know if you have any further feedback.

rodocontegni:
@luke-j thank you so much for the hard work
is there anything holding up this PR? cc @balazsorban44 
i'm happy to take a shot if there's anything pending

balazsorban44:
I'll have a look soon, sorry for the delay. I was getting the core repo ready for stable v4 in the past couple of weeks, but most of the important PRs have been merged, I'll focus on adapters and the docs in the coming weeks!

packages/sequelize/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "next.js",
+    "oauth",
+    "sequelize"
+  ],
+  "private": false,
+  "publishConfig": {
+    "access": "public"
+  },
+  "scripts": {
+    "test": "jest",
```
balazsorban44, line 24: Don't need to run build on test, we will build before publishing anyway.

packages/sequelize/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "build": "tsc"
+  },
+  "files": [
+    "README.md",
+    "dist"
+  ],
+  "peerDependencies": {
+    "next-auth": ">4 || 4.0.0-beta.1 - 4.0.0-beta.x",
```
balazsorban44, line 32: Corrected this as per #274

balazsorban44 (approved):
All tests pass locally, types look great, let's get this in! :+1:

I will appreciate it if @luke-j and @rodocontegni could be around when v4 goes stable (in ~2 weeks) in case an issue arises! :pray:

Nice work and thank you so much for your contribution and patience! :green_heart:

balazsorban44:
Enjoy `@next-auth/sequelize-adapter`

PR #238, opened 2021-09-19
feat(dynamodb): use new Adapter API

safinn:
## Reasoning 💡

Update the DynamoDB adapter to use the V4 adapter API.

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

Beginner with DynamoDB so may not be perfect/correct but would be looking at using this so tried to do the update.
Diff is a little painful. Maybe better to just view the file. Functions of concern:
- `getUserByAccount`
- `linkAccount`
- `unlinkAccount`

### DIFF
--- a/packages/dynamodb/package.json
+++ b/packages/dynamodb/package.json
@@ -30,7 +30,7 @@
   "license": "ISC",
   "peerDependencies": {
     "dynamodb": "^1.3.0",
-    "next-auth": "^3.17.2"
+    "next-auth": "^>=4"
   },
   "devDependencies": {
     "@types/dynamodb": "^1.2.2",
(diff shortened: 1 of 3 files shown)


balazsorban44:
General feedback:



1. try not to create objects with a restrictive number of fields.

Eg. when y return a user, instead of:

 ```js
dbUser // let's say you got this from the db.

return {
  id: dbUser.Item.id,
  name: dbUser.Item.name
}
```

just do:

```js
return dbUser.Item
```

2.
Just drop `createdAt` and `updatedAt` fields. They arent needed in the default adapters.

safinn:
> 1. Doesn't work without setting session.jwt: true

I believe there are implementations for each function in the V4 adapter interface. Not sure why this would be the case.
> 2. Is it supposed to add a new record with a different id, each time I sign in (eg via GitHub)?

I believe with V3 you couldn't login with the same email using a different service but not sure if that is the case with V4.

PR #243, opened 2021-09-19
docs(typeorm): update `README.md`

ndom91:
Update links and references, etc. for adapters@next / next-auth@beta

## Reasoning 💡

Update the `typeorm-legacy` readme with new links, references, updated badges, etc. for the adapters@next and next-auth@beta packages, etc.

## Checklist 🧢

- [x] Documentation
- ~[ ] Tests~
- [x] Ready to be merged

### DIFF
--- a/packages/typeorm-legacy/README.md
+++ b/packages/typeorm-legacy/README.md
@@ -30,20 +30,13 @@ npm install next-auth@beta @next-auth/typeorm-legacy-adapter@next typeorm
 
 ```js
 import NextAuth from "next-auth"
-import GoogleProvider from "next-auth/providers/google"
 import { TypeORMLegacyAdapter } from "@next-auth/typeorm-legacy-adapter"
-
 
 // For more information on each option (and a full list of options) go to
 // https://next-auth.js.org/configuration/options
 export default NextAuth({
   // https://next-auth.js.org/configuration/providers
-  providers: [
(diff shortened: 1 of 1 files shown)


packages/typeorm-legacy/README.md
```diff
@@ -7,43 +7,44 @@
...
 // For more information on each option (and a full list of options) go to
 // https://next-auth.js.org/configuration/options
 export default NextAuth({
   // https://next-auth.js.org/configuration/providers
   providers: [
-    Providers.Google({
+    GoogleProvider({
       clientId: process.env.GOOGLE_ID,
       clientSecret: process.env.GOOGLE_SECRET,
     }),
```
balazsorban44, line 45: I think we should just drop `providers` from READMEs and focus on the `adapter`. Most users probably won't use Google anyway, and the first thing they do after pasting is to remove it.

balazsorban44 (changes requested):

packages/typeorm-legacy/README.md
ndom91: Yeah sure thing, so no providers in the example code in the adapters?

balazsorban44: Yes. Don't think it is necessary

balazsorban44 (approved):

PR #257, opened 2021-09-24
fix(typeorm-legacy): refer to entities by string instead of class

### LINKED ISSUE #228: EntityMetadataNotFoundError when using custom TypeORM model
Trying to rename some columns in the default schema such that they all use snake case. I have the below schema, adjusted from the docs to remap the names to snake case:

```js
import {
  Entity,
  PrimaryGeneratedColumn,
  Column,
  ManyToOne,
  OneToMany,
  ValueTransformer,
} from "typeorm";

const transformer: Record<"date" | "bigint", ValueTransformer> = {
  date: {
    from: (date: string | null) => date && new Date(parseInt(date, 10)),
    to: (date?: Date) => date?.valueOf().toString(),
  },
  bigint: {
    from: (bigInt: string | null) => bigInt && parseInt(bigInt, 10),
    to: (bigInt?: number) => bigInt?.toString(),
  },
};

@Entity({ name: "user" })
export class UserEntity {
  @PrimaryGeneratedColumn("uuid")
  id!: string;

  @Column({ type: "varchar", nullable: true })
  name!: string | null;

  @Column({ type: "varchar", nullable: true, unique: true })
  email!: string | null;

  @Column({ type: "varchar", nullable: true, transformer: transformer.date })
  emailVerified!: string | null;

  @Column({ type: "varchar", nullable: true })
  image!: string | null;

  @OneToMany(() => SessionEntity, (session) => session.userId)
  sessions!: SessionEntity[];

  @OneToMany(() => AccountEntity, (account) => account.userId)
  accounts!: AccountEntity[];
}

@Entity({ name: "account" })
export class AccountEntity {
  @PrimaryGeneratedColumn("uuid")
  id!: string;

  @Column({ type: "uuid" })
  userId!: string;

  @Column({ type: "text" })
  type!: string;

  @Column({ type: "text" })
  provider!: string;

  @Column({ type: "text" })
  providerAccountId!: string;

  @Column({ type: "varchar", nullable: true })
  refresh_token!: string;

  @Column({ type: "varchar", nullable: true })
  access_token!: string | null;

  @Column({
    nullable: true,
    type: "bigint",
    transformer: transformer.bigint,
  })
  expires_at!: number | null;

  @Column({ type: "varchar", nullable: true })
  token_type!: string | null;

  @Column({ type: "varchar", nullable: true })
  scope!: string | null;

  @Column({ type: "varchar", nullable: true })
  id_token!: string | null;

  @Column({ type: "varchar", nullable: true })
  session_state!: string | null;

  @Column({ type: "varchar", nullable: true })
  oauth_token_secret!: string | null;

  @Column({ type: "varchar", nullable: true })
  oauth_token!: string | null;

  @ManyToOne(() => UserEntity, (user) => user.accounts, {
    createForeignKeyConstraints: true,
  })
  user!: UserEntity;
}

@Entity({ name: "session" })
export class SessionEntity {
  @PrimaryGeneratedColumn("uuid")
  id!: string;

  @Column({ type: "text", unique: true })
  sessionToken!: string;

  @Column({ type: "uuid" })
  userId!: string;

  @Column({ transformer: transformer.date, type: "date" })
  expires!: string;

  @ManyToOne(() => UserEntity, (user) => user.sessions)
  user!: UserEntity;
}

@Entity({ name: "verification_token" })
export class VerificationTokenEntity {
  @PrimaryGeneratedColumn("uuid")
  id!: string;

  @Column({ type: "text" })
  token!: string;

  @Column({ type: "text" })
  identifier!: string;

  @Column({ transformer: transformer.date, type: "date" })
  expires!: string;
}

```

However, this gives me `error - EntityMetadataNotFoundError: No metadata for "UserEntity" was found.`. Is there something wrong with the schema or is this a bug with the adapter?

nextauth.ts:
```js
import NextAuth from "next-auth";
import GoogleProvider from "next-auth/providers/google";
import EmailProvider from "next-auth/providers/email";
import { TypeORMLegacyAdapter } from "@next-auth/typeorm-legacy-adapter";
import * as entities from "lib/entities";

export default NextAuth({
  providers: [
    EmailProvider({
      server: process.env.EMAIL_SERVER,
      from: process.env.EMAIL_FROM,
    }),
    GoogleProvider({
      clientId: process.env.GOOGLE_CLIENT_ID,
      clientSecret: process.env.GOOGLE_CLIENT_SECRET,
    }),
  ],
  adapter: TypeORMLegacyAdapter(
    {
      url: process.env.DATABASE_URL,
      type: "postgres",
      ssl: {
        ca: process.env.DATABASE_CERT,
        rejectUnauthorized: process.env.DOPPLER_ENV === "dev" ? false : true,
      },
    },
    { entities }
  ),
  debug: true,
});

```

jrasanen:
## Reasoning 💡

TypeORM doesn't seem to always load entities properly on NextJS.

Also the objects are sometimes not ready after the HMR has been executed.

By referring the entities by class string instead of classes fixes the issue.

I didn't run the tests locally as I couldn't get them working on an M1 mac for some reason, but will try on a Linux later.

Bt the way, is there a reason why `await` the task inside instead of just returning the promise?

## Checklist 🧢

~- [ ] Documentation~
- [x] Tests
- [x] Ready to be merged

## Affected issues 🎟

Fixes #228

### DIFF
--- a/packages/typeorm-legacy/src/index.ts
+++ b/packages/typeorm-legacy/src/index.ts
@@ -64,9 +64,6 @@ export function TypeORMLegacyAdapter(
     },
   }
 
-  const { UserEntity, AccountEntity, SessionEntity, VerificationTokenEntity } =
-    c.entities
-
   return {
     /**
      * Method used in testing. You won't need to call this in your app.
@@ -79,27 +76,27 @@ export function TypeORMLegacyAdapter(
     // @ts-expect-error
     createUser: async (data) => {
       const m = await getManager(c)
-      const user = await m.save(UserEntity, data)
+      const user = await m.save("UserEntity", data)
       return user
     },
     // @ts-expect-error
     async getUser(id) {
       const m = await getManager(c)
-      const user = await m.findOne(UserEntity, { id })
+      const user = await m.findOne("UserEntity", { id })
       if (!user) return null
       return { ...user }
     },
     // @ts-expect-error
     async getUserByEmail(email) {
       const m = await getManager(c)
-      const user = await m.findOne(UserEntity, { email })
+      const user = await m.findOne("UserEntity", { email })
       if (!user) return null
       return { ...user }
     },
     async getUserByAccount(provider_providerAccountId) {
       const m = await getManager(c)
       const account = await m.findOne<Account & { user: AdapterUser }>(
-        AccountEntity,
+        "AccountEntity",
         provider_providerAccountId,
         { relations: ["user"] }
       )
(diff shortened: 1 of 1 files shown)


balazsorban44:
Not awaiting properly throws this ESLint error:

https://github.com/typescript-eslint/typescript-eslint/blob/v4.30.0/packages/eslint-plugin/docs/rules/return-await.md

jrasanen:
> I did not know this was possible! How does typeorm know that the string corresponds to a certain entity though?

> UPDATE: I don't really use classes in JS, but it seems that class Foo {} has its name available as a string like Foo.name // 'Foo' 😅 So that makes sense then. I guess typeorm does a search on name amongst the entities.

@balazsorban44 I believe referring classes using strings in TypeORM was made to allow cyclic relations in database models (don't quote me on that though), [this is the type for](https://github.com/typeorm/typeorm/blob/master/src/common/EntityTarget.ts) `entityClass` which is the first parameter in for most DB operation methods.

> Do you have an error stack or similar? Could we do something about it, or do you think the problem is elsewhere?

I have to get back to this later, can't remember what the issue was: something about node-gyp failing on ARM.

jrasanen:
Hopefully, this would resolve the entities not found errors.

I tested it with our company's client project, and it worked. Getting this merged would allow us to use NextAuth beta in production. 😎

balazsorban44:
I'll have a look very 🔜!

balazsorban44 (approved):
Awesome! All tests pass, and it looks like we don't need to change any docs either!

PR #302, opened 2021-11-08
MySQL Adapter

roelandmoors:
## Reasoning 💡

A new Adaptor for Mysql and PlanetScale

## Checklist 🧢

- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged

This needs more documentation and real life testing, but I'm making this draft PR to get some input. Is this a good start?

## Affected issues 🎟

See [this](https://github.com/nextauthjs/adapters/discussions/259) discussion

### DIFF
--- a/packages/mysql/README.md
+++ b/packages/mysql/README.md
@@ -40,7 +40,7 @@ const connection = mysql.createConnection(config)
 export default NextAuth({
   // https://next-auth.js.org/configuration/providers
   providers: [],
-  adapter: MysqlAdapter(connectionPromise)
+  adapter: MysqlAdapter(connection)
   ...
 })
 ```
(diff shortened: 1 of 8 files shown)


roelandmoors:
I have 2 tests that fail because of a small time difference:
```
    -   "emailVerified": 2021-11-08T19:15:50.901Z,
    +   "emailVerified": 2021-11-08T19:15:50.973Z,
```
Not sure how to solve this?

The expires_at field in the accounts table is a BIGINT. 

What type of fields should I use for this?

roelandmoors:
Not sure why that token is needed? 
I don't think that secret is available with a pull request.

```
      - uses: actions/checkout@v2
        with:
          fetch-depth: 0
          persist-credentials: false
          token: ${{ secrets.GH_PAT }}
```

roelandmoors:
There is already a Prisma and TypeORM adapter so maybe this was not the best idea.

balazsorban44:
We plan to abandon TypeORM for SQL flavor specific adapters in the future. But no problem, we will revisit it anyway.

PR #314, opened 2021-11-20
refactor(neo4j): clean up Neo4j adapter

balazsorban44:
Continues on #186

- extracting `from` and `to` format methods similar to other adapters
- ran Prettier
- simplified queries
- split `query` to `read` and `write` methods
- whenever possible, I drop `RETURN` from the statements, in hope that the writes are faster this way, and it would require less converting using `format`
- remove any restrictions on what data the user can pass (eg.: remove explicit `expires`, `emailVerified` etc. formatting.)

cc @EarthlingDavey

Still have problems with the `.cypher` file, even the CI publishing failed because of it https://github.com/nextauthjs/adapters/runs/4270643543?check_suite_focus=true, so for now, I just ignore it.

### DIFF
--- a/packages/neo4j/src/index.ts
+++ b/packages/neo4j/src/index.ts
@@ -6,55 +6,51 @@ import { client, format } from "./utils"
 export { format }
 
 export function Neo4jAdapter(session: Session): Adapter {
-  const query = client(session)
+  const { read, write } = client(session)
 
   return {
     async createUser(data) {
       const user: any = { id: uuid(), ...data }
-      await query(`CREATE (u:User $u)`, { u: format.to(user) })
+      await write(`CREATE (u:User $data)`, user)
       return user
     },
 
     async getUser(id) {
-      return await query(
-        `MATCH (u:User { id: $id }) RETURN properties(u)`,
-        { id },
-        { tx: "read" }
-      )
+      return await read(`MATCH (u:User { id: $id }) RETURN properties(u)`, {
+        id,
+      })
     },
 
     async getUserByEmail(email) {
-      return await query(
+      return await read(
         `MATCH (u:User { email: $email }) RETURN properties(u)`,
-        { email },
-        { tx: "read" }
+        { email }
       )
     },
 
     async getUserByAccount(provider_providerAccountId) {
-      return await query(
+      return await read(
         `MATCH (u:User)-[:HAS_ACCOUNT]->(a:Account {
            provider: $provider,
            providerAccountId: $providerAccountId
          })
          RETURN properties(u)`,
-        provider_providerAccountId,
-        { tx: "read" }
+        provider_providerAccountId
       )
     },
 
-    async updateUser(user) {
-      return await query(
-        `MATCH (u:User { id: $u.id })
-         SET u += $u
+    async updateUser(data) {
+      return await write(
+        `MATCH (u:User { id: $data.id })
+         SET u += $data
          RETURN properties(u)`,
-        { u: format.to(user) }
+        data
       )
     },
 
     async deleteUser(id) {
-      return await query(
-        `MATCH (u:User { id: $id })
+      return await write(
+        `MATCH (u:User { id: $data.id })
          WITH u, properties(u) AS properties
          DETACH DELETE u
          RETURN properties`,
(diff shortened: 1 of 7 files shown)


packages/neo4j/src/utils.ts
```diff
@@ -1,93 +1,65 @@
...
+  /(\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d:[0-5]\d|Z))|(\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d([+-][0-2]\d:[0-5]\d|Z))|(\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d([+-][0-2]\d:[0-5]\d|Z))/
 
-// Function to transform date values from neo4j DateTime to JS Date object.
-const neo4jDateToJs = (value: typeof neo4j.DateTime | null) => {
-  if (!value || !neo4j.temporal.isDateTime(value)) {
-    return value
-  }
-  return new Date(value.toString())
+function isDate(value: any) {
+  return value && isoDateRE.test(value) && !isNaN(Date.parse(value))
```
balazsorban44, line 9: This is the only part I did not really like, but I cannot see another way of solving this.

Neo4j does not accept `Date`, so they are saved as ISO strings.

When reading the values ideally these would be converted back to `Date`, but Neo4j cannot do that. We have to either detect ISO Date strings somehow. This is the safest way I could do that.

packages/neo4j/src/index.ts
```diff
@@ -1,224 +1,165 @@
...
-        }
+    async linkAccount(data) {
+      const account = { id: uuid(), ...data }
+      await write(
+        `MATCH (u:User { id: $data.userId })
+         MERGE (a:Account {
+           providerAccountId: $data.providerAccountId,
+           provider: $data.provider
+         }) 
+         ON CREATE SET a.id = $data.id
```
EarthlingDavey, line 69: With this change, userId is set on the Account node. 
I think this should is not necessary, because here we create a relationship between Account and User. And, this relationship is be used in `getUserByAccount`.

packages/neo4j/src/index.ts
```diff
@@ -1,224 +1,165 @@
...
-          RETURN a { .*, userId: u.id } AS a`,
-        {
-          id: uuid(),
-          userId,
-          providerAccountId,
-          provider,
-          accountData,
-        }
+    async linkAccount(data) {
+      const account = { id: uuid(), ...data }
```
EarthlingDavey, line 62: I think `id: uuid()` isn't neded here.

EarthlingDavey, line 62: I think one line is incorrect here, either
- `id: uuid()` isn't neded here .
- or `ON CREATE SET a.id = $data.id` should be `ON CREATE SET a.id = $id`

EarthlingDavey:
Thanks for taking time to do this! Really great to learn from these changes. I've left 2 comments, but I have a 2yo child on my lap 😂 I may have missed something. I'll run test locally later to see it in action 😎

I'm abit lost about the cypher file. I'll ask neo4j communnity.

packages/neo4j/src/index.ts
```diff
@@ -1,224 +1,165 @@
...
+           providerAccountId: $data.providerAccountId,
+           provider: $data.provider
+         }) 
+         ON CREATE SET a.id = $data.id
+         SET a += $data
+         MERGE (u)-[:HAS_ACCOUNT]->(a)`,
+        account
       )
+      return account
     },
```
EarthlingDavey, line 75: I think one line is incorrect round here, either
- `id: uuid()` isn't neded here .
- or later `ON CREATE SET a.id = $data.id` should be `ON CREATE SET a.id = $id`

packages/neo4j/src/index.ts
```diff
@@ -1,224 +1,165 @@
...
-          MERGE (u)-[:HAS_ACCOUNT]->(a)
-          RETURN a { .*, userId: u.id } AS a`,
-        {
-          id: uuid(),
-          userId,
-          providerAccountId,
-          provider,
-          accountData,
-        }
+    async linkAccount(data) {
```
EarthlingDavey, line 61: I think one line is incorrect round here, either
- `id: uuid()` isn't neded here .
- or later `ON CREATE SET a.id = $data.id` should be `ON CREATE SET a.id = $id`

packages/neo4j/src/index.ts
```diff
@@ -1,224 +1,165 @@
...
+      await write(
+        `MATCH (u:User { id: $data.userId })
+         MERGE (a:Account {
+           providerAccountId: $data.providerAccountId,
+           provider: $data.provider
+         }) 
+         ON CREATE SET a.id = $data.id
+         SET a += $data
+         MERGE (u)-[:HAS_ACCOUNT]->(a)`,
+        account
```
EarthlingDavey, line 72: I think one line is incorrect round here, either
- `id: uuid()` isn't neded here .
- or later `ON CREATE SET a.id = $data.id` should be `ON CREATE SET a.id = $id`

balazsorban44:
thanks, I'll visit your comments today. All tests pass for me locally, so I felt kind of confident with my changes, but obviously I haven't used Neo4j myself before, so the extra pair of eyes 👀 help a lot!

moving the constraints setup into the test itself would be amazing! check the `connect` method of the `basicTests` function, very similar to `disconnect`, what you already used for the watch part.

ndom91 (approved):
All tests pass for me as well, nice clean up with the format date to/from methods. This is how we did it for the other adapters as well, as each one seems to have its own quirks in regard to storing and returning datetime column types.

PR #316, opened 2021-11-22
chore(neo4j): minor amend to tests

EarthlingDavey:
## Reasoning 💡

After a read through @balazsorban44 's refactor I made a few minor amends. Thanks btw, I leadned a lot!

- Add to tests to fail if userId foreign key exists on session or account.
- CreateSession, don't add userId on Session node
- For the above change to CreateSession, amend format.to to allow nested objects
- On create statements, = can be uses instead of +=

## Checklist 🧢

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged

## Affected issues 🎟

Following on from PR https://github.com/nextauthjs/adapters/pull/314

### DIFF
--- a/packages/neo4j/src/utils.ts
+++ b/packages/neo4j/src/utils.ts
@@ -16,6 +16,7 @@ export const format = {
     for (const key in object) {
       const value = object[key]
       if (value instanceof Date) newObject[key] = value.toISOString()
+      else if (value instanceof Object) newObject[key] = format.to(value)
       else newObject[key] = value
     }
     return newObject
--- a/packages/neo4j/src/utils.ts
+++ b/packages/neo4j/src/utils.ts
@@ -16,7 +16,6 @@ export const format = {
     for (const key in object) {
       const value = object[key]
       if (value instanceof Date) newObject[key] = value.toISOString()
-      else if (value instanceof Object) newObject[key] = format.to(value)
       else newObject[key] = value
     }
     return newObject
--- a/packages/neo4j/tests/index.test.ts
+++ b/packages/neo4j/tests/index.test.ts
@@ -42,7 +42,7 @@ runBasicTests({
       const session = result?.records[0]?.get("s")?.properties
       const userId = result?.records[0]?.get("userId")
 
-      if (!session || !userId) return null
+      if (!session || session.userId || !userId) return null
 
       return { ...format.from(session), userId }
     },
@@ -59,7 +59,8 @@ runBasicTests({
       const account = result?.records[0]?.get("a")?.properties
       const userId = result?.records[0]?.get("userId")
 
-      if (!account || !userId) return null
+      if (!account || account.userId || !userId) return null
+
       return { ...format.from(account), userId }
     },
 
@@ -76,3 +77,4 @@ runBasicTests({
     },
   },
 })
+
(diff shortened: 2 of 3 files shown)


packages/neo4j/tests/index.test.ts
```diff
@@ -42,7 +42,7 @@ runBasicTests({
       const session = result?.records[0]?.get("s")?.properties
       const userId = result?.records[0]?.get("userId")
 
-      if (!session || !userId) return null
+      if (!session || session.userId || !userId) return null
```
balazsorban44, line 45: Why is this necessary?

packages/neo4j/tests/index.test.ts
```diff
@@ -59,7 +59,8 @@ runBasicTests({
       const account = result?.records[0]?.get("a")?.properties
       const userId = result?.records[0]?.get("userId")
 
-      if (!account || !userId) return null
+      if (!account || account.userId || !userId) return null
```
balazsorban44, line 62: Why is this necessary?

packages/neo4j/src/utils.ts
```diff
@@ -16,6 +16,7 @@ export const format = {
     for (const key in object) {
       const value = object[key]
       if (value instanceof Date) newObject[key] = value.toISOString()
+      else if (value instanceof Object) newObject[key] = format.to(value)
```
balazsorban44, line 19: I think this would be better handled where needed in the adapter. just split the stuff that is an object and run formatting on it.

balazsorban44 (changes requested):
Left

packages/neo4j/tests/index.test.ts
EarthlingDavey: I think this and the `|| account.userId` are needed to catch the case where the userId property has been accidentally set by the adapter. For example, if the test was like this during your refactor it might have helped you catch the change that I made in this PR re: `async createSession(data) {`.

There is nothing to say that another developer might come along and get hit by the same gotcha. I mean, adding all the data to the Session node does look like a logical way to clean up the code. This clause just gaurds against it.

packages/neo4j/src/utils.ts
EarthlingDavey: Thanks for the suggestion. It's addressed here: https://github.com/nextauthjs/adapters/pull/316/commits/2b2f588af32820eedd2277cf73ac8d23492e116d

packages/neo4j/tests/index.test.ts
balazsorban44: It sounds unnecessary to guard against that at runtime for every user. It did not produce an error in the tests, so technically this is not even a problem IMO.

If you think the tests should catch this, maybe we should add some extra tests at the bottom of the test file instead.

EarthlingDavey: This is the tests file so it doesn't run at runtime. I did think about additonal tests but it's a bunch more LOC for something pretty simple.

balazsorban44: My bad... :sweat_smile:

balazsorban44 (approved):
Thank you!

EarthlingDavey:
No problem. Thanks for the review. Cheers

PR #317, opened 2021-11-22
fix(dgraph): refactor, simplify, fix tests

balazsorban44:
Continues on #206.

@arnaudDerbey,@lveillard, @bmartel please have a look, feel free to comment!

Firstly, I removed unnecessary fields. `createdAt`, `updatedAt` isn't necessary as I mentioned earlier https://github.com/nextauthjs/adapters/pull/206#pullrequestreview-811935525

`VerificationRequest` is a remainder of the old database schemas, and should be called `VerificationToken` now. The adapter is merely a broker between the core and the database, it does not perform any other requests (it was the case before, that the adapter sent the email as well)

I removed the queries and mutations files. They are completely unnecessary. Most adapters fit into a single file, even with their verbose query language. I think it is much easier to understand what the adapter does, than jumping between files to get the context. Feel free to disagree on this one.

I expanded on the Fragments. In general, we should not limit any of the models, so **all 4** models should be extendable.

Removed the `class` for the dgraph client. It really can be just an object of the configuration for the client, similar to the TypeORM Adapter.

The actual client should not care about the fragments either, it is just the client that makes the requests as expected. fragments config moved into the adapter, check its new signature.

One less import, and hopefully easier to understand.

The tests were unfortunately giving false positives. The tests used the adapter methods to test against, while that's exactly what we are trying to test. They should be raw queries to the database. I could not find some of these in the queries file I removed, so I started this PR by removing the false tests and I will add them back later on.

Furthermore, we should **not** query against a database over the network. It is slow, and unreliable. I also found a secret in the test file, this should be invalidated immediately! :warning:

We need a way to run against a database locally for the tests to be more reliable.

Did some general cleanup as well, removed self-explanatory comments, `console.log`s, etc.

All of this will have to be updated in the docs as well. https://github.com/nextauthjs/docs/pull/120 is not merged yet, so feel free to update that.


balazsorban44:
User and Session tests should be fixed, working on the Account one...

I really don't like when backends take and transform data without a word:

image

We send in a number, and Dgraph just decides to save it as a Date anyway... I will now have to find a way to prevent this. :confused:

balazsorban44:
Until we find a way to run tests locally, how do I fix the schema @arnaudDerbey,@lveillard, @bmartel?

Getting these type of errors:

```sh
 Cannot query field "addVerificationToken" on type "Mutation". Did you mean "addVerificationRequest", "updateVerificationRequest", or "deleteVerificationRequest"?
    Unknown type "AddVerificationTokenInput"
```

balazsorban44:
I'll see if we can use this https://hub.docker.com/r/dgraph/dgraph to create a local server to test against, similar to the other adapters

balazsorban44:
So this https://github.com/nextauthjs/adapters/pull/317#issuecomment-975947064 happened because the schema defined `expires_at` as `DateTime`, which is wrong. It is supposed to be a number, so I fixed it in the schema

balazsorban44:
Fixed all the tests, and using Docker, I created a local instance to run the tests against. I think it is ready for review.

balazsorban44:
Documentation should be updated accordingly

arnaudDerbey:
Awesome work @balazsorban44 !!

balazsorban44:
I feel confident, but feel free to comment on it if you think I made a mistake or something unusual for the dgraph world.

arnaudDerbey:
@balazsorban44 regarding your question about getUserByEmail that return an array. It's a mistake I've made because my knowledge was limited.  
The User type inside the schema  should define the email field as an id. (with the @id directive) This would define the uniqueness of that field and allow us to queryUser by email without filtering like we do today and get a unique user. 
I think there are more fields that should be redefined with the @id directive. 

I think we shouldn't push the PR before we improve the schema.

packages/dgraph/src/client.ts
```diff
@@ -0,0 +1,58 @@
+import * as jwt from "jsonwebtoken"
+import fetch, { HeadersInit } from "node-fetch"
+
+export interface DgraphClientParams {
+  endpoint: string
```
bmartel, line 5: I believe for Dgraph cloud there needs to be an option for an apiKey which can be set https://dgraph.io/docs/cloud/admin/authentication/. What you have here looks great for self managed and local deploys of DGraph but without the client apiKey option which would be set under the header `Dg-Auth: <your-api-key>' ` this will not work for those folks using DGraph cloud as the queries made by non authenticated users (ie: login, register etc.) won't be available to anonymous users of a secured schema.

bmartel: I don't believe the current implementation is secure https://github.com/nextauthjs/adapters/blob/next/packages/dgraph/src/graphql/secure.schema.gql#L44. Can someone explain how `$nextAuth` protects the queries/mutations, and who sets this value?

bmartel:
> @bmartel could you update the corresponding docs PR once you are satisfied with how these changes look?
@balazsorban44 Things look really good, I have a couple questions I posed about the Dgraph cloud usage, and security. Beyond that I think this is good to go. Upon getting some clarity on those two questions I posed, I will be updating the docs to reflect everything here.

packages/dgraph/src/client.ts
bmartel: For reference I typically treat the "system" or "admin" as I refer to it as with a separate secret that gets injected into the schema which looks like this:

```graphql
type Account
  @auth(
    delete: { rule: "{$adminSecret: { eq: \"<ADMIN_SECRET>\" } }" }
    add: { rule: "{$adminSecret: { eq: \"<ADMIN_SECRET>\" } }" }
    query: { rule: "{$adminSecret: { eq: \"<ADMIN_SECRET>\" } }" }
    update: { rule: "{$adminSecret: { eq: \"<ADMIN_SECRET>\" } }" }
  ) {}
``
and declare the env in `pages/api/[...nextauth]`
```tsx
export default NextAuth({
  adapter: DgraphAdapter({
    endpoint: process.env.NEXT_PUBLIC_GRAPHQL_URL!,
    apiKey: process.env.GRAPHQL_API_KEY,
    adminSecret: process.env.ADMIN_SECRET,
    authHeader: process.env.AUTH_HEADER,
    jwtSecret: process.env.JWT_SECRET,
  }),
})
```

That way there is a secure comparison between a client request and the graphql api in Dgraph. As far as I can see, `$nextAuth` will always resolve true and allow any request access to all values in the secure schema, rendering it the same as the unsecure variant.

balazsorban44:
I think I'll refactor this code:
https://github.com/nextauthjs/adapters/blob/4c9b4376189e4aa60869537b10208506ab166d4f/packages/dgraph/src/client.ts#L4-L10

From my understanding, the `authHeader` option is unnecessary, as it can be `Authorization` https://dgraph.io/docs/cloud/cloud-api/overview/#understanding-headers-used-across-the-apis

Then if you set the `jwtSecret`, then we append that header to the requests.

packages/dgraph/src/client.ts
balazsorban44: Here: https://github.com/nextauthjs/adapters/blob/4c9b4376189e4aa60869537b10208506ab166d4f/packages/dgraph/src/client.ts#L35-L37

We generate the content of the JWT that is sent to Dgraph. In our case, the content is `nextAuth: true`, which signals Dgraph that we have the correct "role" to execute operations like this:

https://github.com/nextauthjs/adapters/blob/4c9b4376189e4aa60869537b10208506ab166d4f/packages/dgraph/src/graphql/secure.schema.gql#L3

bmartel: Just read through this prior to you answering, retracted those statements as this is correct. Awesome :+1:

bmartel: I think as long as we can verify this configuration is valid for DGraph cloud, this is great.

bmartel: docs updated nextauthjs/docs#120

balazsorban44:
I had some problems working out how to use `RS256` locally. I will try to get it working, as it seems to be a more secure alternative than `HS256`

https://community.auth0.com/t/rs256-vs-hs256-jwt-signing-algorithms/58609

bmartel:
> I had some problems working out how to use `RS256` locally. I will try to get it working, as it seems to be a more secure alternative than `HS256`
> 
> https://community.auth0.com/t/rs256-vs-hs256-jwt-signing-algorithms/58609

Definitely is a better default for sure. What end of things were causing issues when switching to RS256?

balazsorban44:
just setting it up locally. I'll have to try later, but feel free to check out the branch and try yourself, it's appreciated!

bmartel:
> just setting it up locally. I'll have to try later, but feel free to check out the branch and try yourself, it's appreciated!

I have a PR against this one that outlines the changes required, I used the basis tests to verify this. #320

balazsorban44:
@bmartel thank you so much, it helped a lot! I know refactored the tests to only run with RSA256. I don't think we want to advocate for anything else anyway, so it should be fine.

PR #322, opened 2021-11-25
feat: add collectionPrefix config option to MongoDB adapter

4nduril:
## Reasoning 💡

This adds a configuration option to the new MongoDB adapter:
The option is called `collectionPrefix` and enables the user to provide a prefix that is used for all database collection names that the adapter uses.

This was possible in next-auth v3 via the `entityPrefix` option for TypeORM.

## Checklist 🧢

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## Affected issues 🎟

Closes #321

PR #323, opened 2021-11-25
feat(mongodb): simplify adapter initialization

### LINKED ISSUE #321: Make it possible to use prefix for collections
### Title

Make it possible to use prefix for collections like entityPrefix in TypeORM

### How to reproduce ☕️

I don't think it's useful to provide code to reproduce a feature request here.

### Your question/bug report 📓

With next-auth v3 I used the typeORM adapter with MongoDB as shown in the docs. There was also a hint (https://next-auth.js.org/v3/configuration/databases#how-to-use-a-database) how you could use `entityPrefix` to have all "tables" aka collections that the adapter is going to create get a prefixed name.

Today I switched to v4 and adapters@next and as per documentation I used the separate adapter for MongoDB. But this one is lacking the option for collection name prefixing. 

My suggestion would be an additional config option besides `db` which would be optional and called `collectionPrefix`. 

I had a quick look on the code of the adapter and it's so straight forward that I would implement it and open a PR against the next branch if that's fine? Any thoughts?

### Contributing  🙌🏽

Yes, I am

balazsorban44:
Looking through issues, the initialization of the MongoDB adapter has been a pain for newcomers and those used to the other adapters. I've found a way to match the other adapters' initialization process, while only requiring that the user adds the database name on their connection string.

This PR also closes #322, closes #321

Docs changes at https://github.com/nextauthjs/docs/pull/133

The new syntax is:
```js
import NextAuth from "next-auth"
import { MongoDBAdapter } from "@next-auth/mongodb-adapter"
import clientPromise from "lib/mongodb"

export default NextAuth({
  adapter: MongoDBAdapter(clientPromise),
  // https://next-auth.js.org/configuration/providers
  providers: [],
  ...
})
```

You can also pass a second parameter, that looks like this:

```ts
interface MongoDBAdapterOptions {
  collections?: {
    Users?: string
    Accounts?: string
    Sessions?: string
    VerificationTokens?: string
  }
}
```

### DIFF
--- a/packages/mongodb/README.md
+++ b/packages/mongodb/README.md
@@ -73,8 +73,6 @@ import clientPromise from "lib/mongodb"
 // https://next-auth.js.org/configuration/options
 export default NextAuth({
   adapter: MongoDBAdapter(clientPromise),
-  // https://next-auth.js.org/configuration/providers
-  providers: [],
   ...
 })
 ```
--- a/packages/mongodb/README.md
+++ b/packages/mongodb/README.md
@@ -1,7 +1,7 @@
 <p align="center">
    <br/>
    <a href="https://next-auth.js.org" target="_blank"><img height="64px" src="https://next-auth.js.org/img/logo/logo-sm.png" /></a>&nbsp;&nbsp;&nbsp;&nbsp;<img height="64px" src="./logo.svg" />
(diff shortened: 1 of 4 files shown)


ndom91 (approved):
Nice find! The good ol double await haha

LGTM :+1:

PR #325, opened 2021-11-26
feat(mikro-orm): add mikro-orm adapter

boredland:
## Reasoning 💡

Mainly personal reasons, [mikro orm](https://github.com/mikro-orm/mikro-orm) seems to be a well documented full-featured ORM with unit of work / repository patterns that deals pretty well with the somewhat chaotic sql ecosystem and is pretty extendable in that direction (typed json fields, custom fieldtypes). Went with the upcoming mikro-orm v5, since it fixes a bug around cascade delete in sqlite and has even more typesafety (typesafe .toObject). And its release is eminent.

## Checklist 🧢

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

~I'd like a close look from @balazsorban44 regarding the many type-casts I had to make to comply to the "Adapter" interface. According to him the generic Adapter-Interface "was unnecessary", so I'm not sure I'm doing things right there.~


balazsorban44 (changes requested):
Nice work, all tests pass already! To make this on par with the other adapters and make people transfer easier, I would like if the public API is as similar as possible. This requires hiding the connection part.

I recently did something very similar with the MongoDB adapter at #323

Could you refactor the initialization to match that? TypeORM is also very similar to that https://github.com/nextauthjs/adapters/blob/next/packages/typeorm-legacy/src/index.ts

Maybe the init options could be the first parameter of the adapter, very similar to the TypeORM one. the second parameter could be the options, with an optional entities object, where you can override the models.

packages/mikro-orm/src/models.ts
```diff
@@ -0,0 +1,162 @@
+import { randomUUID } from "crypto"
```
balazsorban44, line 1: This would only work in Node 14 and upwards. I think for now using `uuid` as an extra dependency is a way to go.

packages/mikro-orm/src/index.ts
```diff
@@ -0,0 +1,148 @@
+import { RequestContext, wrap } from "@mikro-orm/core"
+
+import {
+  Account,
+  Session,
+  User as DefaultUserModel,
+  VerificationToken,
+} from "./models"
```
balazsorban44, line 8:
```suggestion
import * as defaultModels from "./models"
```

I would use a * import and re-export the models as well to the user, in case they could use it for something else.

packages/mikro-orm/src/index.ts
```diff
@@ -0,0 +1,148 @@
...
+import {
+  Adapter,
+  AdapterSession,
+  VerificationToken as AdapterVerificationToken,
+} from "next-auth/adapters"
+import { DefaultAccount } from "next-auth"
```
balazsorban44, line 14:
```suggestion
import type {
  Adapter,
  AdapterSession,
  VerificationToken as AdapterVerificationToken,
} from "next-auth/adapters"
import type { DefaultAccount } from "next-auth"
```

Let's mark type imports with the `type` keyword for clarity.

This is true for all the places, but I will only comment here.

boredland: Hi! I'd introduce a linting rule for that. But I already saw that not all adapter comply to it, so I wont make it an Error for now.

boredland:
> I would like if the public API is as similar as possible. This requires hiding the connection part.

Would passing in an instance of EntityManager be ok for you or is each Adapter supposed to work completely without an outside dependency? As I see it, the idea of a [Request Context](https://mikro-orm.io/docs/next/identity-map/#-requestcontext-helper-for-di-containers) would be hard to align with the adapter initializing on its own (which may not be too bad after al..)

balazsorban44:
I want the user to pass us an instance of any kind, so we can rely on their ORM as a peer-dependency. Check out the TypeORM adapter, which I think is pretty similar to what I would like to see.

boredland:
I honestly don't exactly know whats happening in that typeORM adapter, especially around that `__disconnect` - at least I fear that is the open handles I have when passing in the mikro-orm instance as a promise. Maybe you have an idea?

boredland:
I reverted back to allowing to pass in an optional Entity Manager as a Promise. In contexts like an API route, MikroORM should be provided with a RequestContext ("... RequestContext helper, that will use node's AsyncLocalStorage in the background to isolate the request context."). The RequestContext now can either be retrieved from outside the Next-Auth handler (like [here](https://github.com/jonahallibone/mikro-orm-nextjs/blob/master/pages/api/add.ts#L9)). I kept this optional, since there is no real difference to doing this inside the Next-Auth route imo (and as a bonus I don't have the open handles issue inside the text context).

balazsorban44:
AsyncLocalStorage has only been stable since Node 16. :thinking: If that's the case, it should be documented.

Could you start working on the docs page maybe at https://github.com/nextauthjs/docs?

balazsorban44 (changes requested):
We would like to release a new version of `next-auth` v4 in a few days, so to get this in before that, I committed most of my suggestions.

The README.md needs a refactor, and please open a corresponding PR at https://github.com/nextauthjs/docs as well. :pray: 

I based some of my suggestions off of the TypeORM adapter.

packages/mikro-orm/README.md
```diff
@@ -0,0 +1,77 @@
...
+         global.__MikroORM__ = await MikroORM.init(config)
+      }
+      return global.__MikroORM__;
+   };
+
+   // ensures mikro-orm RequestContext exists on the api route
+   const withORM = (handler: NextApiHandler) => async (req: NextApiRequest, res: NextApiResponse) => {
+      const orm = await getORM();
+      return RequestContext.createAsync(orm.em, async () => handler(req, res));
+   }
```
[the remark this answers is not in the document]
boredland: I understand your concerns with the RequestContext, especially due to it basing on a NodeJS 16 Feature. I have not been aware of that. I reverted back to the version with just an a promise for the orm, people would have to pass an and updated the readme.

packages/mikro-orm/README.md
```diff
@@ -0,0 +1,71 @@
...
+            debug: process.env.DEBUG === "true" || process.env.DEBUG?.includes("db"),
+         })
+      }
+      return global.__MikroORM__
+   }
+
+
+   // For more information on each option (and a full list of options) go to
+   // https://next-auth.js.org/configuration/options
+   export default NextAuth(req, res, {
```
balazsorban44, line 56:
```suggestion
   import { MikroOrmAdapter } from "@next-auth/mikro-orm-adapter"

   // fetches a global instance of mikro-rom
   const getORM = async () => {
      if (!global.__MikroORM__) {
         global.__MikroORM__ = await MikroORM.init({
            dbName: "./db.sqlite",
            type: "sqlite",
            entities: [User, Session, Account, VerificationToken],
            debug: process.env.DEBUG === "true" || process.env.DEBUG?.includes("db"),
         })
      }
      return global.__MikroORM__
   }


   // For more information on each option (and a full list of options) go to
   // https://next-auth.js.org/configuration/options
   export default NextAuth({
```

balazsorban44: Remember, we are trying to do this as simple as possible. It shouldn't be necessary by default to import anything else than the adapter itself.

What `mikro-orm` seems to be doing reminds me of this Prisma suggestion very much btw: https://www.prisma.io/docs/support/help-articles/nextjs-prisma-client-dev-practices#solution

boredland:
> so to get this in before that, I committed most of my suggestions.

perhaps I am misunderstanding sth, but I did't see commits or suggestions by you? Anyhow: I reverted back to the orm-promise. But I still have the openhandles problem - perhaps you have a solution for that?

I am open to create a PR against docs, but would like to wait until you "ok" this one here (so I don't have to adapt them too often).

balazsorban44 (changes requested):

packages/mikro-orm/README.md
boredland: I think that kind of approach is pretty common in the nextjs world, especially with vercel, because the "global" scope is often "warmed" up on consecutive requests and thus your connection might already be initialized.

balazsorban44: Still seeing the extra imports from the adapter. Since we know the default entities, isn't there a way to just let the user not pass those at init?

I want the initialization to be as simple as this, or as close as we can get: 
image

balazsorban44: My suggestion commits haven't been pushed, and now there are conflicts... :sweat_smile: I'll try to fix them and push

boredland: ah, damn.

balazsorban44:
Please check out my changes, and update the README.md accordingly. Try getting rid of the need for the user to pass the entities. Otherwise, there is almost no point in providing the built-in ones. :shrug:

boredland:
I think the only option for users not to pass the entities is, if we initialize mikro-orm inside the adapter. Which we now would do, if a user is not passing in Mikro-Orm Options instead of the Promise.

packages/mikro-orm/src/entities.ts
```diff
@@ -0,0 +1,146 @@
...
+    hidden: true,
+    orphanRemoval: true,
+    cascade: [Cascade.ALL],
+  })
+  accounts = new Collection<Account>(this)
+}
+
+@Entity()
+export class Session implements AdapterSession {
+  @PrimaryKey({ default: randomUUID() })
```
B4nan, line 66: this is wrong, the `default` option in decorator is for database defaults, here you would give it a specific UUID generated on require time

packages/mikro-orm/src/entities.ts
```diff
@@ -0,0 +1,146 @@
...
+    cascade: [Cascade.ALL],
+  })
+  sessions = new Collection<Session>(this)
+
+  @OneToMany({
+    entity: () => Account,
+    mappedBy: (account) => account.user,
+    hidden: true,
+    orphanRemoval: true,
+    cascade: [Cascade.ALL],
```
B4nan, line 59: `orphanRemoval` implies `Cascade.REMOVE`, this line is a no-op

B4nan (changes requested):
two small nits

packages/mikro-orm/src/entities.ts
boredland: I think this is a debugging / try-around leftover when it didn't work with sqlite - good catch.

packages/mikro-orm/src/entities.ts
boredland: same here, thanks!

packages/mikro-orm/src/index.ts
```diff
@@ -0,0 +1,210 @@
...
+export * as defaultEntities from "./entities"
+
+/**
+ * The MikroORM adapter accepts a MikroORM instance or configuration and returns a NextAuth adapter.
+ * @param ormConnection can either be an instance promise or a MikroORM connection configuration (https://mikro-orm.io/docs/next/configuration#driver)
+ * @param options entities in the options object will be passed to the MikroORM init function as entities
+ * @returns
+ */
+export function MikroOrmAdapter<
+  TUser extends typeof defaultEntities.User,
```
balazsorban44, line 17: I would still prefer not to expose any generics. Isn't there like a general `Entity` type that would make TS happy on the `entities` option? I don't want the user to have to configure types on the adapter. It doesn't mean that much to me if its not perfect internally, tests should be added to cover custom entities instead.

balazsorban44 (changes requested):
Looks almost good! Either way, I'll merge this later today if the tests pass, the user facing API looks good to me now!

packages/mikro-orm/src/index.ts
boredland: I think as its now, it is unnecessary anyway. Will try and drop it.

balazsorban44 (approved):
Nice work, thank you so much! :green_heart:

PR #329, opened 2021-12-01
refactor(dynamo): clean up

balazsorban44:
Some last-minute cleanup of the source-code @mathisobadia, hope you don't mind. It's mostly simplified to match the style of the other adapters as well.


mathisobadia:
@balazsorban44 of course no problem, consistency is good! ;) I didn't take the time to check other adapters to see how they were made when I first opened the PR because it was not clear to me which ones were fully updated into the new api and which one were not but good that will all have the same quality standard now!

PR #333, opened 2021-12-03
fix(typeorm): do not parse date strings in `format`

rozenmd:
## Reasoning 💡

As far as I can tell (let me know if i'm wrong!), TypeORM returns timestamps as strings, not integers.

Without this change, SessionEntity returns the following:

```
{
  sessionAndUser: SessionEntity {
    id: 47,
    sessionToken: 'xxxxx',
    userId: 1,
    expires: Invalid Date,
    user: UserEntity {
      id: 1,
      name: 'Max Rozen',
      email: 'xxxxx',
      emailVerified: Invalid Date,
      image: 'xxxxx'
    }
  }
}
```
Which throws a SESSION_ERROR.

With this change, it returns:
```
{
  sessionAndUser: SessionEntity {
    id: 47,
    sessionToken: 'xxxxx',
    userId: 1,
    expires: 2022-01-02T07:52:58.960Z,
    user: UserEntity {
      id: 1,
      name: 'Max Rozen',
      email: 'xxxxx',
      emailVerified: 2021-07-12T22:21:36.302Z,
      image: 'xxxxx'
    }
  }
}
```

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/packages/typeorm-legacy/src/entities.ts
+++ b/packages/typeorm-legacy/src/entities.ts
@@ -9,7 +9,7 @@ import {
 
 const transformer: Record<"date" | "bigint", ValueTransformer> = {
   date: {
-    from: (date: string | null) => date && new Date(parseInt(date, 10)),
+    from: (date: string | null) => date && new Date(date),
     to: (date?: Date) => date?.valueOf().toString(),
   },
   bigint: {


rozenmd:
Actually @balazsorban44 - it might just be a wrong assumption on my part, not a bug in the adapter...

In my DB, `expires` and `emailVerified` are stored as `timestamp with timezone`, as a result the existing transformers break login and new account creation:

```js
{
    from: (date: string | null) => date && new Date(parseInt(date, 10)),
    to: (date?: Date) => date?.valueOf().toString(),
  }
```

when I change them to just return the date, everything works:

```js
{
    from: (date: string | null) => date,
    to: (date?: Date) => date,
  }
```

which to me implies the transformers for dates aren't necessary - unless in next-auth v4 dates aren't dates any more?

balazsorban44:
Adapters should return `Date` in any case. Could you run the tests locally to verify? I don't remember from the top of my head. :sweat_smile: 

if you have a different schema than the default one though, you might have to customize the entities sent into the adapter. https://next-auth.js.org/adapters/typeorm#custom-models

rozenmd:
Ah yeah, my PR breaks the tests. 

TypeORM expects the date fields to be varchar instead of timestamp so my PR will break things. Will just use a custom model instead.

PR #336, opened 2021-12-05
Allow larger id_token in typeorm adapter

### LINKED ISSUE #230: Typeorm-legacy. Login with Google Account: ER_DATA_TOO_LONG: Data too long for column 'id_token' at row
Issue regarding the typeorm-legacy default entity model

This is the error I got. 
The  entity for this database field "id_token" is a varchar, which is capped on 255 characters.

This happens when:
When trying to sign in with my Google Account, the id_token is 1054 characters.

So the type of that field should be "text" or equivalent. I am not great at all when it comes to databases.

### LINKED ISSUE #3825: Typeorm-legacy. Login with Google Account: ER_DATA_TOO_LONG: Data too long for column 'id_token' at row

thgh:
## Reasoning 💡

This resolved my issue with sign in with Google using TypeORM and seems applicable to everyone.

This change raises the max size from 255 to 65k characters.

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

## Affected issues 🎟

Resolves nextauthjs/next-auth#3825

### DIFF
--- a/packages/typeorm-legacy/src/entities.ts
+++ b/packages/typeorm-legacy/src/entities.ts
@@ -78,7 +78,7 @@ export class AccountEntity {
   @Column({ type: "varchar", nullable: true })
   scope!: string | null
 
-  @Column({ type: "varchar", nullable: true })
+  @Column({ type: "text", nullable: true })
   id_token!: string | null
 
   @Column({ type: "varchar", nullable: true })


thgh:
I think TypeORM can change the column type without losing data. But I'm using Prisma now, so not sure about it!

PR #341, opened 2021-12-10
feat(upstash-redis): add upstash-redis adapter

kay-is:
## Reasoning 💡

I want to add a new adapter for the Upstash Redis service.

## Checklist 🧢

- [x] Documentation
- [x] Tests
- [ ] Ready to be merged

Please let me know what additional steps are required to merge. Like, where to store the Upstash API keys for the tests?

Thank you!


packages/upstash-redis/package-lock.json
```diff
@@ -0,0 +1,69 @@
+{
```
balazsorban44, line 1: This file should be removed. We use `yarn` and `lerna` in the repository.

balazsorban44 (commented):
Just curious, what's Upstash about this adapter? Could this work with any Redis client? :thinking: For testing purposes, it would be ideal if we had a local client set up with Docker if needed, just as the other adapters.

kay-is:
Thanks for the quick reply!

The Upstash Redis client uses a connectionless HTTP-based protocol, which allows deployment on more systems than a regular Redis client. For example, Cloudflare Workers only support outgoing HTTP.

A regular Redis client would work with non-Upstash Redis, but couldn't be deployed on systems that only support HTTP connections.

balazsorban44:
So there is no way to connect to a local upstash instance for testing purposes then? 🤔

ndom91:
> So there is no way to connect to a local upstash instance for testing purposes then? thinking
> 
> cc @ndom91

So I couldn't find any docker container or first party offline dev solution, but would it maybe make sense to run it in cloudflare's local worker env (`wrangler dev`)? I guess it would still need to connect to the upstash cloud in that case, right? Defeating the purpose of local cloudflare for testing :thinking:

kay-is:
@ndom91 The idea is to use a free tier account for the tests.

ndom91:
So currently we've explicitly gone with a local variant of every database, even cloud based ones like fauna, for testing. @balazsorban44 what do you think? Would it be okay to use a free cloud account for testing this one?

balazsorban44:
Sounds like there is no workaround, so we should create a free-tier database in Upstash, and embed the secret in the GitHub Actions environment. This means we might not be able to run tests on fork PRs for this Adapter - something that doesn't work right now anyway because of how we rely on some secrets to check the changes in a PR - not now, not in the future.

If it's something we accept, then I guess this is alright.

ndom91:
> Sounds like there is no workaround, so we should create a free-tier database in Upstash, and embed the secret in the GitHub Actions environment. This means we might not be able to run tests on fork PRs for this Adapter - something that doesn't work right now anyway because of how we rely on some secrets to check the changes in a PR - not now, not in the future.
> 
> If it's something we accept, then I guess this is alright.

Yeah I mean its up to you if we accept it or not haha. But I don't see another option either, and as long as they provide at least a free dev account we can use I don't think its that big of a deal. As you mentioned, it might be difficult with forks due to sharing of repo secrets, but thats an issue for another day :sweat_smile:

@kay-is sounds good to me then!

kay-is:
Cool, thank you for being so chill about it!

A new Upstash account can be created [here](https://console.upstash.com/login).

Should I change the code to use ENV variables in the tests
and you set up the API keys then via GH Actions?

ndom91:
> Cool, thank you for being so chill about it!
> 
> A new Upstash account can be created [here](https://console.upstash.com/login).
> 
> Should I change the code to use ENV variables in the tests and you set up the API keys then via GH Actions?

Yeah so as we mentioned, the secrets don't get shared with forks, but I've added the secrets to the main `nextauthjs/adapters` repo.

1. `NEXTAUTH_UPSTASH_REDIS_URL`
2. `NEXTAUTH_UPSTASH_REDIS_KEY`

balazsorban44:
I'll try to review this in the holidays (probably next week), feel free to ping me in early next year if I haven't responded :pray:

kay-is:
@balazsorban44 just a review reminder 🔍

balazsorban44:
Tahnk you, I'll schedule this for this weekend!

packages/upstash-redis/tests/index.test.ts
```diff
@@ -0,0 +1,39 @@
+import upstashRedisClient from "@upstash/redis"
+import { runBasicTests } from "../../../basic-tests"
+import UpstashRedisAdapter from "../src"
+
+const client = upstashRedisClient(
+  process.env.NEXTAUTH_UPSTASH_REDIS_URL,
+  process.env.NEXTAUTH_UPSTASH_REDIS_KEY
```
balazsorban44, line 7:
```suggestion
  process.env.UPSTASH_REDIS_URL,
  process.env.UPSTASH_REDIS_KEY
```

I feel this prefix is unnecessary in this context.

packages/upstash-redis/tests/index.test.ts
```diff
@@ -0,0 +1,39 @@
+import upstashRedisClient from "@upstash/redis"
+import { runBasicTests } from "../../../basic-tests"
+import UpstashRedisAdapter from "../src"
+
```
balazsorban44, line 4:
```suggestion
import "dotenv/config"

```

packages/upstash-redis/tests/index.test.ts
```diff
@@ -0,0 +1,39 @@
...
+const client = upstashRedisClient(
+  process.env.NEXTAUTH_UPSTASH_REDIS_URL,
+  process.env.NEXTAUTH_UPSTASH_REDIS_KEY
+)
+
+const reviveFromJson = (json: string) =>
+  JSON.parse(json, (key, value) =>
+    key === "emailVerified" || key === "expires" ? new Date(value) : value
+  )
```
balazsorban44, line 13: This should probably come from the adapter

packages/upstash-redis/src/index.ts
```diff
@@ -0,0 +1,209 @@
...
+
+  const getUser = async (id: string) => {
+    const response = await client.get(userKeyPrefix + id)
+    if (!response.data) return null
+    return reviveFromJson(response.data)
+  }
+
+  const adapter: Adapter = {
+    createUser: async (user: Omit<AdapterUser, "id">): Promise<AdapterUser> => {
+      const id = uuid.v4()
```
balazsorban44, line 102:
```suggestion
      const id = uuid()
```

packages/upstash-redis/src/index.ts
```diff
@@ -0,0 +1,209 @@
+import type { Account as AdapterAccount } from "next-auth"
+import type {
+  Adapter,
+  AdapterUser,
+  AdapterSession,
+  VerificationToken,
+} from "next-auth/adapters"
+import type { Upstash } from "@upstash/redis/src/types"
+import * as uuid from "uuid"
```
balazsorban44, line 9:
```suggestion
import { v4 as uuid } from "uuid"
```

packages/upstash-redis/src/index.ts
```diff
@@ -0,0 +1,209 @@
...
+  accountKeyPrefix?: string
+  accountByUserIdPrefix?: string
+  emailKeyPrefix?: string
+  sessionKeyPrefix?: string
+  sessionByUserIdKeyPrefix?: string
+  userKeyPrefix?: string
+  verificationTokenKeyPrefix?: string
+}
+
+const defaultOptions = {
```
balazsorban44, line 22:
```suggestion
export const defaultOptions = {
```
Let's make it easy to reuse if necessary.

packages/upstash-redis/src/index.ts
```diff
@@ -0,0 +1,209 @@
...
+  const sessionByUserIdKeyPrefix =
+    baseKeyPrefix + mergedOptions.sessionByUserIdKeyPrefix
+  const userKeyPrefix = baseKeyPrefix + mergedOptions.userKeyPrefix
+  const verificationTokenKeyPrefix =
+    baseKeyPrefix + mergedOptions.verificationTokenKeyPrefix
+
+  const reviveFromJson = (json: string) =>
+    JSON.parse(json, (key, value) =>
+      key === "emailVerified" || key === "expires" ? new Date(value) : value
+    )
```
balazsorban44, line 57: I generally don't like it if Date fields are converted by their name. It makes it harder if the user would like to add another Date field. They'll need to modify the adapter.

This seems to be a reoccurring thing with adapters. So far very few adapters respect JS `Date` as-is. One of them is the Prisma adapter.

I think we will need to expose a transform configuration option from adapters. Feel free to ignore this comment for now, it's more of a future self-note.

packages/upstash-redis/src/index.ts
```diff
@@ -0,0 +1,209 @@
...
+    return user
+  }
+
+  const getUser = async (id: string) => {
+    const response = await client.get(userKeyPrefix + id)
+    if (!response.data) return null
+    return reviveFromJson(response.data)
+  }
+
+  const adapter: Adapter = {
```
balazsorban44, line 100:
```suggestion
  const adapter: Adapter = {
```

this is unnecessary. just return the adapter object directly. You will be able to drop all the argument typings.

Please also use the `async methodName() {}` format for consistency with other adapters.

packages/upstash-redis/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+```js
+npm install next-auth @next-auth/upstash-redis-adapter @upstash/redis
+```
+
+2. Add the follwing code to your `pages/api/[...nextauth].js` next-auth configuration object.
+
+```js
+import NextAuth from "next-auth"
+import UpstashRedisAdapter from "@next-auth/upstash-adapter"
```
balazsorban44, line 31:
```suggestion
import { UpstashRedisAdapter } from "@next-auth/upstash-adapter"
```

Prefer named imports

packages/upstash-redis/src/index.ts
```diff
@@ -0,0 +1,209 @@
...
+  accountKeyPrefix: "user:account:",
+  accountByUserIdPrefix: "user:account:by-user-id:",
+  emailKeyPrefix: "user:email:",
+  sessionKeyPrefix: "user:session:",
+  sessionByUserIdKeyPrefix: "user:session:by-user-id:",
+  userKeyPrefix: "user:",
+  verificationTokenKeyPrefix: "user:token:",
+}
+
+export default function UpstashRedisAdapter(
```
balazsorban44, line 33:
```suggestion
export function UpstashRedisAdapter(
```

Prefer named export

packages/upstash-redis/tests/index.test.ts
```diff
@@ -0,0 +1,39 @@
...
+  adapter: UpstashRedisAdapter(client, { baseKeyPrefix: "testApp:" }),
+  db: {
+    verificationToken: async (where) => {
+      const { data } = await client.get(
+        `testApp:user:token:${where.identifier}`
+      )
+      return reviveFromJson(data)
+    },
+    user: async (id: string) => {
+      const { data } = await client.get(`testApp:user:${id}`)
```
balazsorban44, line 25:
```suggestion
      const { data } = await client.get(`testApp:user:${id}`)
```

Why are `testApp` prefixes necessary, if you already added a `baseKeyPrefix` on line 16?

packages/upstash-redis/tests/index.test.ts
```diff
@@ -0,0 +1,39 @@
...
+
+const reviveFromJson = (json: string) =>
+  JSON.parse(json, (key, value) =>
+    key === "emailVerified" || key === "expires" ? new Date(value) : value
+  )
+
+runBasicTests({
+  adapter: UpstashRedisAdapter(client, { baseKeyPrefix: "testApp:" }),
+  db: {
+    verificationToken: async (where) => {
```
balazsorban44, line 18: I would prefer the `async verificationToken(where)() {}` format in the test file as well.

Furthermore, could we change the order to user, account, session, verificationToken?

balazsorban44 (changes requested):
Thank you! I left some comments. Mostly formatting, the rest looks good!

balazsorban44:
I'm curious if we can utilize https://docs.upstash.com/redis/features/restapi#pipelining somehow to reduce the number of requests where there are multiple get or set?

kay-is:
@balazsorban44 thanks for the review! 

I applied your edits in my last commit.

packages/upstash-redis/package-lock.json
balazsorban44: @kay-is, it looks like this file was not removed. Could you remove it, please?

packages/upstash-redis/package.json
```diff
@@ -0,0 +1,43 @@
...
+  ],
+  "peerDependencies": {
+    "next-auth": "^4.0.1",
+    "@upstash/redis": "^0.2.1"
+  },
+  "devDependencies": {
+    "@upstash/redis": "^0.2.1"
+  },
+  "dependencies": {
+    "dotenv": "^10.0.0",
```
balazsorban44, line 40: This should be in `devDependencies`, as it is only used for testing purposes.

packages/upstash-redis/src/index.ts
```diff
@@ -0,0 +1,212 @@
...
+    },
+    async linkAccount(account: AdapterAccount): Promise<AdapterAccount> {
+      const id = `${account.provider}:${account.providerAccountId}`
+      return await setAccount(id, { ...account, id })
+    },
+    async createSession(session: {
+      sessionToken: string
+      userId: string
+      expires: Date
+    }): Promise<AdapterSession> {
```
balazsorban44, line 140:
```suggestion
    async createSession(session) {
```

Typings like these on the returned adapter methods should not be necessary, they are inferred from the `Adapter` interface added on line 45. Could you remove them?

balazsorban44 (changes requested):
Thank you, almost good to go :slightly_smiling_face:, a few suggestions left.

Could you also open a PR at the docs? 
https://github.com/nextauthjs/docs/tree/main/docs/adapters
https://github.com/nextauthjs/docs/blob/main/docs/adapters/overview.md
https://github.com/nextauthjs/docs/blob/main/sidebars.js#L56

We might also need an issue template and enable auto labeling:

https://github.com/nextauthjs/adapters/tree/main/.github/ISSUE_TEMPLATE
https://github.com/nextauthjs/adapters/blob/main/.github/labeler.yml

balazsorban44:
I'm curious, should we just call this `@next-auth/upstash-adapter` instead, especially if we don't plan to make it a general-purpose Redis adapter anyway?

kay-is:
@balazsorban44 I updated the code again and made a PR for the docs. I also updated the GitHub issue files.

The name "upstash-redis" is to make sure it's for Upstash's Redis offering because they offer multiple databases. I also wanted to clarify it's not a generic Redis client.

balazsorban44:
Hmm. What other databases are they offering? I only see Kafka, but I don't know if that is similar to traditional databases? Am I wrong?

kay-is:
If I understood it correctly, their USP is to offer serverless versions to well-known databases. They started with Redis, then Kafka. 

They might add Elasticsearch or Postgress in the future. I'm only a freelancer tasked to write this adapter, so I have no more profound insight into their internals, sorry.

balazsorban44:
Thank you. It's probably fine to be future-safe then.

balazsorban44 (changes requested):
I removed the `package-lock.json` file myself, seems to have been forgotten.

But could you please run `yarn` in the root folder and commit the `yarn.lock` file change after that?

kay-is:
I updated the lock file.

It was a pleasure working with you; I even learned something about JSON date-parsing and TypeScript!

balazsorban44 (approved):
Thank you!

PR #343, opened 2021-12-13
[prisma_adapter] delete session catch P2025

identiq:
## Reasoning 💡

When a user has a sessionToken (stored in browser cookies), and missing from the database sessions, he can't connect anymore 
The adapter wasn't catching error P2025 on deleteSession as for VerificationToken

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/packages/prisma/src/index.ts
+++ b/packages/prisma/src/index.ts
@@ -30,8 +30,17 @@ export function PrismaAdapter(p: PrismaClient): Adapter {
     createSession: (data) => p.session.create({ data }),
     updateSession: (data) =>
       p.session.update({ data, where: { sessionToken: data.sessionToken } }),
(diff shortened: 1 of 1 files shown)


balazsorban44:
I wonder in which case would that happen? 🤔 Whenever we check a session and it is invalid, we clean up both the database and the cookie.

identiq:
@balazsorban44 In my case, I changed cookies configuration for sub domains isolation

willemmulder:
I just had this happen to me when I switched our database URL. Thanks for this bug report, because it was the only thing I could find, and clearing cookies did the trick as a workaround!

balazsorban44:
This repository has been merged into a monorepo. See https://github.com/nextauthjs/next-auth instead.

PR #348, opened 2021-12-20
fix(neo4j): update user bugfix

EarthlingDavey:
## Reasoning 💡

Using the neo4j adapter, I noticed that if I sign in with Facebook, then, with email - it fails. 

The reason being that `updateUser` returned an object like `{ 'properties(u): { id: '123', name: 'abc' } }`,}`

This PR fixes that by returning a user object from`updateUser`

In addition I have used shorter notation `u{.*}` instead of `properties(u)` throughout all neo4j statements.

And, to catch this bug in future, I have made an edit to basic-tests.js for your consideration.

## Checklist 🧢

~- [ ] Documentation~
- [x] Tests
- [x] Ready to be merged


balazsorban44 (approved):

balazsorban44:
@EarthlingDavey could you reopen this against `main`? We are out of beta. :)

EarthlingDavey:
Sure thing: https://github.com/nextauthjs/adapters/pull/352

PR #355, opened 2021-12-27
feat(mongoose): New Mongoose adapter

choutkamartin:
## Reasoning 💡
I want to add a new adapter for Mongoose (object modelling tool for MongoDB). Current MongoDB adapter cannot be used with an active Mongoose connection.

## Checklist 🧢

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged

## Affected issues 🎟
None


choutkamartin:
As already mentioned in #294 

Hey, I have created a Mongoose adapter for MongoDB. The current MongoDB adapter doesn't support an already active Mongoose connection.

### Tests
All tests are passing.

I tried the adapter in my test application. Everything works just fine.

### Documentation
I can write the documentation in no time - if this PR will get merged.

### Additional things to look out for
I'm not sure if the current way how the adapter is connected to the database works flawlessly. Yes, the tests are passing, but I think the connection should be somehow cached, like in [Next example](https://github.com/vercel/next.js/blob/canary/examples/with-mongodb-mongoose/lib/dbConnect.js) or as already used in the MongoDB adapter. Because right now, I get a whopping 23 connections on a localhost development server with a remote MongoDB server. Seems like the connections don't get closed, and they are not cached, so every time I start the development server, another connection is initialized.

choutkamartin:
I figured that using a Mongoose adapter isn't the best idea, here's why:
### The problem
The adapter connects through the `mongoose.connect()` method, which establishes a connection to the entire project. That can be handy, but there's a drawback. The models in the adapter are bound to the default connection, which means that other models the user creates will collide with the existing ones from the adapter.

For example:
The user will create a model `User` in his project, like this:
```js
import mongoose from "mongoose";

const userSchema = new mongoose.Schema({
  email: {
    type: String,
    unique: true,
    required: true,
  },
  name: {
    type: String,
  },
  password: {
    type: String, // Notice the password property
    required: true,
  },
});
export default mongoose.models.User || mongoose.model("User", userSchema);
```

Take a look at the password property. This property isn't present in the model of the adapter. To create a new document the user can do this:
```js
const user = new User({
    email: email@email.com,
    password: "password",
    name: "Test",
});
```
This should create a new document in the database, with all three properties present. But because the password property isn't present in the adapter model, the property will get omitted. That's a major issue. The `User` model from the adapter and the user-created one collide.
### What can be done?
What can be done, is to create separate connections for the adapter itself. There's a big difference between `mongoose.connect()` and `mongoose.createConnection()` methods. While the first one creates only one connection for the entire project, with the second one we can assign connections to specific models. So, we can do something like this: 
```js
// index.test.ts
import { runBasicTests } from "../../../basic-tests"
import { MongooseAdapter } from "../src"
import { createConnection } from "mongoose"
import {
  userSchema,
  accountSchema,
  sessionSchema,
  verificationTokenSchema,
} from "../src/models"

const uri: string = "mongodb://localhost:27017/test"

var conn = createConnection(uri)

const UserModel = conn.model("User", userSchema);
const AccountModel = conn.model("Account", accountSchema);
const SessionModel = conn.model("Session", sessionSchema);
const VerificationTokenModel = conn.model("VerificationToken", verificationTokenSchema);

runBasicTests({
  adapter: MongooseAdapter(conn),
  db: {
  },
})
```
```js
// index.ts
/* eslint-disable @typescript-eslint/no-non-null-assertion */
import {
  userSchema,
  accountSchema,
  sessionSchema,
  verificationTokenSchema,
} from "./models"
import type {
  Adapter,
} from "next-auth/adapters"
import { Connection } from "mongoose"

export function MongooseAdapter(conn: Connection): Adapter {
  const UserModel = conn.model("User", userSchema);
  const AccountModel = conn.model("Account", accountSchema);
  const SessionModel = conn.model("Session", sessionSchema);
  const VerificationTokenModel = conn.model("VerificationToken", verificationTokenSchema);

  return {
  }
}
```
This way, we are saying, hey, we want to connect these specific models to the database and don't use this connection for anything else.

Now, if the adapter uses a separate connection, the end-user can do this in his project.
```js
await connect(process.env.MONGODB_URI);
```
With this method, other models in his project will get connected, yet, the adapter models will use their assigned connection.

If we will do what we wanted before (creating a user with password property), we will see, that the property will be saved to the DB just fine.

Now, we use two connections to the database (pooling excluded).

### Why in the end creating a Mongoose adapter isn't useful
Now with all these troubles, I don't think there's a point in having a separate adapter for Mongoose. If someone will use Mongoose in a project, they should use a MongoDB adapter, create a simple connection like this (below) and use a connection for their Mongoose models in a separate middleware.
```js
import NextAuth from "next-auth";
import { MongoDBAdapter } from "@next-auth/mongodb-adapter";
import { MongoClient } from "mongodb";

const uri = process.env.MONGODB_URI;
const client = new MongoClient(uri);
const clientPromise = client.connect();

export default NextAuth({
  adapter: MongoDBAdapter(clientPromise),
  providers: [
  ],
});
```

It would be a good idea to notice this in the MongoDB documentation. Someone may try to use Mongoose connection with the MongoDB adapter and wonder why it doesn't work.

AdaptingAFM:
> Now with all these troubles, I don't think there's a point in having a separate adapter for Mongoose. If someone will use Mongoose in a project, they should use a MongoDB adapter, create a simple connection like this (below) and use a connection for their Mongoose models in a separate middleware.

@choutkamartin 

How would you then go about operating with any `User` document inserted by NextAuth with a mongoose connection later on? There's no defined Mongoose schema to interact with.

choutkamartin:
@AdaptingAFM You would create your schema according to the models provided here: https://next-auth.js.org/adapters/models. And then, you could extend the schema by any fields you want.

AnishDe12020:
Just asking, why isn't the `dbConnect` not an asynchronous function?

packages/mongoose/src/dbConnect.ts
```diff
@@ -0,0 +1,13 @@
+import type { Connection } from "mongoose"
+import { createConnection } from "mongoose"
+
+let conn: Connection
+
+export default function dbConnect(uri: string) {
```
AnishDe12020, line 6: Question: Why is this not asynchronous?

packages/mongoose/src/index.ts
```diff
@@ -0,0 +1,146 @@
...
+    async createVerificationToken(data) {
+      const verificationToken = await VerificationTokenModel.create(data)
+      return from<VerificationToken>(verificationToken)
+    },
+    async useVerificationToken(data) {
+      const verificationToken = await VerificationTokenModel.findOneAndDelete(
+        data
+      ).lean()
+      if (!verificationToken) return null
+      const { _id, __v, ...rest } = verificationToken
```
AnishDe12020, line 142: Question: We can use the `from` function from `format` here right? Or else, there will be 2 unused variables

packages/mongoose/src/index.ts
```diff
@@ -0,0 +1,146 @@
...
+  VerificationToken,
+} from "next-auth/adapters"
+import type { Account } from "next-auth"
+import dbConnect from "./dbConnect"
+
+export const format = {
+  /** Takes a mongoDB object and returns a plain old JavaScript object */
+  from<T = Record<string, unknown>>(object: Record<string, any>): T {
+    const newObject: Record<string, unknown> = {}
+    for (const key in object) {
```
AnishDe12020, line 22: I had got eslint complain about the usage of for..in here. `Object.keys(object).forEach` should be fine right?

AnishDe12020 (commented):
I was using some of this code to make a custom adapter (I am using mongoose and going to be adding some more fields to user). Got some errors from my diagonistic tools (eslint and typescript  server) so just want some clarification

PR #356, opened 2021-12-28
Fix issues with firebase adapter

srijans38:
Hi!

I have done three things here:
1. Cleaned up the FirebaseClient interface. It feels very unnecessary to have things like collection, query etc. in there. I believe this just makes it easier to use the adapter.
2. Fix the undefined issue as mentioned and suggested in https://github.com/nextauthjs/adapters/pull/183#discussion_r720880253
3. Fix the tests to support the changed FirebaseClient interface.

Tests are passing, and this should be now usable with Firebase v9.

### DIFF
--- a/packages/firebase/src/index.ts
+++ b/packages/firebase/src/index.ts
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
-import type {
+import {
   runTransaction,
   collection,
   query,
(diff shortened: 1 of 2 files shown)


greendesertsnow:
Hey! Did you test for Firebase v9?

srijans38:
Yeah! This should work for v9. This branch was already working, except for that issue with Google Provider.

viperfx:
Is there a way to use firebase for just authentication without using the firestore and other parts?

PR #363, opened 2022-01-04
feat(firebase) : Add firebase auth support

srijans38:
## Changes

1. Switched to "firebase-admin" from "firebase"
2. Add firebase auth support for creating, updating and deleting a user
3. Add tests for changes

## Testing 

Tested in firebase emulator. firestore working as before. Firebase auth user is created, updated and deleted. 

Also tested in prod with google provider.

### DIFF
--- a/packages/firebase/jest.config.js
+++ b/packages/firebase/jest.config.js
@@ -1,1 +1,6 @@
-module.exports = require("../../jest.config")
+module.exports = {
+  preset: "ts-jest",
+  testEnvironment: "node",
+  modulePathIgnorePatterns: ["dist/", "node_modules/"],
+  resolver: "jest-node-exports-resolver",
+}
--- a/packages/firebase/jest.config.js
+++ b/packages/firebase/jest.config.js
@@ -1,6 +1,6 @@
+const jestConfig = require("../../jest.config")
+
 module.exports = {
-  preset: "ts-jest",
-  testEnvironment: "node",
-  modulePathIgnorePatterns: ["dist/", "node_modules/"],
+  ...jestConfig,
   resolver: "jest-node-exports-resolver",
 }
--- a/packages/firebase/firebase.json
+++ b/packages/firebase/firebase.json
@@ -2,6 +2,9 @@
   "emulators": {
     "firestore": {
       "port": 8080
+    },
+    "auth": {
+      "port": 9099
     }
   }
 }
--- a/packages/firebase/package.json
+++ b/packages/firebase/package.json
@@ -32,11 +32,12 @@
     "test": "FIRESTORE_EMULATOR_HOST=localhost:8080 firebase emulators:exec --only firestore --project next-auth-test jest"
   },
   "peerDependencies": {
-    "firebase": "^9.0.0-beta.8",
+    "firebase-admin": "^10.0.1",
     "next-auth": ">=4"
   },
   "devDependencies": {
-    "firebase": "^9.0.0-beta.8",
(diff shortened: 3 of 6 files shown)


cripsito:
Hello! i'm trying to use the latest changes on the adapter, how can I npm your changes and import them into my project?

srijans38:
@cripsito I really don't have an idea about that. I have used "file:" syntax to test locally built modules, so something like npm install file:<path-to-module> but I am not sure it'll work. This is also a complete refactor so I really wouldn't advise using this in the current state for something serious.

packages/firebase/jest.config.js
```diff
@@ -1 +1,6 @@
-module.exports = require("../../jest.config")
+module.exports = {
+  preset: "ts-jest",
+  testEnvironment: "node",
+  modulePathIgnorePatterns: ["dist/", "node_modules/"],
+  resolver: "jest-node-exports-resolver",
+}
```
ndom91, line 6: Please continue to require the root `jest.config.js`, you can extend it wiht your resolver if needed in this case like this:

```
const jestConfig = require('../../jest.config')

module.exports = {
  ...jestConfig,
  resolver: "jest-node-exports-resolver",
}
```

ndom91:
@cripsito you can install from a specific Github repo / branch like so:

```
npm install "https://github.com/srijans38.git#feat/refactor-firebase" --save
```

ndom91:
Hey @srijans38 just so I understand this correctly, you've added support for firebase auths "custom auth provider" (i.e. like described here - https://firebase.google.com/docs/auth/web/custom-auth) alongside the existing firestore db calls?

So basically the user would be using two firebase products doing the same thing? Have I understood this correctly? I'm no firebase expert :sweat_smile: 

If thats correct, what about adding the firebase-auth as a new separate adapter. That way users can still use firestore as a traditional DB backend for persisting users. But the new adapter then would allow them to use only firebase-auth as a backend for next-auth?

srijans38:
@srijans38 No, using firebase-admin to create users programmatically (https://firebase.google.com/docs/auth/admin/manage-users). I also think these should be seperate adapters this is the issue I raised here (https://github.com/nextauthjs/adapters/issues/180#issuecomment-1004126562. But I don't think it would be possible to implement all the methods required by the adapter api with just firebase auth.

One way I can think of right now is making firebase auth an opt-in option.

ndom91:
Ah thats right, I remember thsi discussion now too, thanks.

So what is the point of adding firebase auth support here exactly? Sorry i'm just not seeing how this improves the firestore-only adapter that already exists. I guess the users would gain some better management functionality of users in the firebase auth UI if the users were created there, instead of just as rows in a firestore db. Is that it? Sorry don't mean to be obtuse, I'm just not following 100% yet haha.

srijans38:
Yeah! I also don't really see a good use case for this. I did it as a proof of concept(kinda) as mentioned here https://github.com/nextauthjs/adapters/issues/180#issuecomment-917615216. Just trying to get a stable release for the adapter. This branch already works with firebase v9 without this pr. If there's anything more to do with #183 , I am more than willing to help.

srijans38:
Yeah! I also don't really see a good use case for this. I did it as a proof of concept(kinda) as mentioned here https://github.com/nextauthjs/adapters/issues/180#issuecomment-917615216. Just trying to get a stable release for the adapter. This branch already works with firebase v9 without this pr. If there's anything more to do with #183 , I am more than willing to help.

PR #365, opened 2022-01-05
Convert value prop to Date only if of type string

### LINKED ISSUE #364: Fauna Ref fields are converted to a JS Date
### Title

Fauna Ref fields are converted to a JS Date

### How to reproduce ☕️

When the user's document contains a field of type Fauna Ref, it is converted to a JS Date by the [format.from](https://github.com/nextauthjs/adapters/blob/50bae0dbb5c85ac88b4c4ce4fb1dc20aef793876/packages/fauna/src/index.ts#L68) function of the `fauna-adapter.`

Here's an example of a user with a Ref field:

```js
{
  ...
  myField: Ref(Collection('myCollection'), '12345'))
}
```

And the result by the `fauna-adapter` when fetching this user: 

```js
{
  ...
  myField: Invalid Date
}
```

It happens because the [format.from](https://github.com/nextauthjs/adapters/blob/50bae0dbb5c85ac88b4c4ce4fb1dc20aef793876/packages/fauna/src/index.ts#L72) function assumes that any field from a fauna object which contains a `value` property is a date, which is partially correct. Indeed, a Ref field returned by fauna also has a `value` property which contains the `id` value.

To fix this issue, I propose to add an extra check inside the `format.from` function at [line 72](https://github.com/nextauthjs/adapters/blob/50bae0dbb5c85ac88b4c4ce4fb1dc20aef793876/packages/fauna/src/index.ts#L72) to make sure the `value` property of the field is of type `string` before converting it to a JS Date; otherwise, we just return the field as it is.

```js
if (value?.value && typeof value.value === 'string') {
  newObject[key] = new Date(value.value)
} else {
  newObject[key] = value
}
```

Please let me know how to proceed with this fix.

### Your question/bug report 📓

Please, see the explanation above.

### Contributing  🙌🏽

Yes, I am

gdangelo:
## Reasoning 💡

Changes to fix #364 on the fauna adapter.

## Checklist 🧢

- ~[ ] Documentation~
- [x] Tests
- [x] Ready to be merged

## Affected issues 🎟

Fixes #364

### DIFF
--- a/packages/fauna/src/index.ts
+++ b/packages/fauna/src/index.ts
@@ -69,7 +69,7 @@ export const format = {
     const newObject: Record<string, unknown> = {}
     for (const key in object) {
       const value = object[key]
-      if (value?.value) {
+      if (value?.value && typeof value.value === 'string') {
         newObject[key] = new Date(value.value)
       } else {
         newObject[key] = value


balazsorban44 (approved):
Thank you!

balazsorban44:
@gdangelo I'm not sure why, this hangs running the tests in the fauna adapter package, and I wasn't able to publish this yesterday. :thinking: 

When I checkout the `main` branch and try to run tests for fauna, it hangs similarly to GitHub Actions. Any idea?

gdangelo:
@balazsorban44 Not sure what's happening. What do you mean by "this hangs running the tests"? Are the tests not running at all? 

I'm trying to run the fauna adapter tests locally, but I got a validation error (see screenshot). Do you know if it's a misconfiguration on my side?

balazsorban44:
We use `lerna`, so make sure you use `yarn` and run `yarn install` in the root folder first.

Tests seem to start, but there is no output, and it never stops.

gdangelo:
Thanks, @balazsorban44. I finally had the tests running locally. All the tests passed successfully (see screenshot) with my changes. Unfortunately, I cannot reproduce the behavior you've described. Do you have any logs or anything we could check to investigate this further?

Screenshot 2022-01-08 at 09 19 23

balazsorban44:
Interesting... All I have is the failing GitHub action

https://github.com/nextauthjs/adapters/actions/runs/1661052430

when I tried locally, it ended up with the same issue.

I'll have to look more into this.

gdangelo:
@balazsorban44 Actually, running the tests from the fauna package was successful. However, running them from the root folder with `lerna run test --since HEAD~` hangs as you said. From what I see, the fauna DB instance starts successfully in the Docker container, but then, nothing...  🤔

Screenshot 2022-01-08 at 15 34 29

balazsorban44:
Yeah. Very curious.

Issue #252, opened 2021-09-20
DynamoDB adapter should use TTL


geovanisouza92:
## Context

Currently the only way to expire _verification tokens_ and _sessions_ is waiting until one of them get read and [check](https://github.com/nextauthjs/adapters/blob/26cd24a6eba3d42ed7febd5eb45b13c236c57819/packages/dynamodb/src/index.ts#L261) if it's [expired](https://github.com/nextauthjs/adapters/blob/26cd24a6eba3d42ed7febd5eb45b13c236c57819/packages/dynamodb/src/index.ts#L415).

This is not a good design, because DynamoDB have the [TTL feature](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/TTL.html) for scenarios like this one.

## Proposed change

- Change the [`createSession`](https://github.com/nextauthjs/adapters/blob/26cd24a6eba3d42ed7febd5eb45b13c236c57819/packages/dynamodb/src/index.ts#L234) method to include an attribute `ttl` with the unix epoch representation of the `expires` attribute.
- Change the [`createVerificationRequest`](https://github.com/nextauthjs/adapters/blob/26cd24a6eba3d42ed7febd5eb45b13c236c57819/packages/dynamodb/src/index.ts#L385) method to include an attribute `ttl` with the unix epoch representation of the `expires` attribute.
- Update the documentation about the possibility, so users of the adapter could configure TTL on their tables.
- Optional: Make the `ttl` attribute name configurable.

anthonyringoet:
I think this can be closed as https://github.com/nextauthjs/adapters/pull/268 was merged.

PR #366, opened 2022-01-05
fix(dynamo): Ensure expires attribute is stored as a UNIX timestamp

GiacoCorsiglia:
## Reasoning 💡

Hi!  This is a small change compared to the number of words and tests I wrote!

This change applies to the DynamoDB adapter.

DynamoDB supports a "TTL" feature that automatically purges old items from the database if they are marked with a special attribute.  The [existing adapter docs](https://next-auth.js.org/adapters/dynamodb) suggest enabling this feature and setting it to use the `expires` attribute.  This is a reasonable choice, because Next Auth already uses the `expires` property on Sessions and VerificationTokens for basically the same purpose.

However, as clarified in the [AWS docs](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/time-to-live-ttl-before-you-start.html#time-to-live-ttl-before-you-start-formatting), the special TTL attribute must be stored as a UNIX timestamp in seconds.  I have fixed a bug in the DynamoDB adapter, which resulted in the `expires` attribute being stored as an ISO string, meaning that the TTL feature would not work for Next Auth Sessions and VerificationTokens.  Following this PR, the TTL feature should work, because now `expires` is stored as a numeric timestamp.

I added some unit tests for the code that converts objects to and from the format saved in DynamoDB.  Most of the tests were passing before this PR, but I thought it was worth adding them anyway.  All the tests pass on my machine with these changes.

The code before I edited it appeared to convert `expires` between a `number` and a `string`.  However, so far as I can tell, the `expires` property on `AdapterSession` and `VerificationToken` is a `Date`, so I think that code path was just being skipped entirely.  I can confirm that, in my database, the `expires` attribute is in fact being stored as an ISO string.  Paging @mathisobadia in case they know what the code was originally trying to do.

Thanks :)

## Checklist 🧢

- ~[ ] Documentation~
- [x] Tests
- [x] Ready to be merged

## Affected issues 🎟

This is related to #252, but that issue was closed (possibly incorrectly? I don't know if this worked in a previous version!).

### DIFF
--- a/packages/dynamodb/src/utils.ts
+++ b/packages/dynamodb/src/utils.ts
@@ -11,10 +11,11 @@ export const format = {
     const newObject: Record<string, unknown> = {}
     for (const key in object) {
       const value = object[key]
-      if (value instanceof Date) newObject[key] = value.toISOString()
-      // Converted to a string by `format.from
-      else if (key === "expires") newObject[key] = parseInt(value, 10)
-      else newObject[key] = value
+      if (value instanceof Date) {
+        // DynamoDB requires the TTL attribute be a UNIX timestamp (in secs).
+        if (key === "expires") newObject[key] = value.getTime() / 1000
+        else newObject[key] = value.toISOString()
+      } else newObject[key] = value
     }
     return newObject
   },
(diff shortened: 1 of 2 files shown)


mathisobadia:
@GiacoCorsiglia Thank you for this! I didn't properly test this functionality on the final version that was merged so that's on me. Looking at my database this worked on the first version I submitted as PR but at some point in the refactoring it stopped working and the elements do have isoString dates now instead of UNIX timestamps. So adding the tests here is a good thing!

GiacoCorsiglia:
Thank you @balazsorban44!

balazsorban44:
I merged, but something broke and the automatic release process is broken in GitHub Actions :confused:

So this has not been released yet, unfortunately. I'm not sure what happened yet, but for some reason, #365 made the Fauna tests hang, both in the CI and on my computer.

balazsorban44:
@GiacoCorsiglia I just saw, this was opened against the `next` branch! Could you please open the same PR against `main` instead?

GiacoCorsiglia:
Done in #367, sorry about that!

(You might want to update [CONTRIBUTING.md](https://github.com/nextauthjs/adapters/blob/main/CONTRIBUTING.md#pull-requests) with whatever branch you prefer folks use 🙂)

PR #375, opened 2022-01-18
fix(sequelize): prevent indexes duplication

na-ji:
## Reasoning 💡

When using the sequelize adapter and running sequelize sync, the unique indexes are duplicated on each run because
they weren't declared correctly. After some time, this error is thrown: `SequelizeDatabaseError: Too many keys specified; max 64 keys allowed`. The fix is to name the indexes to prevent the sync function to give them a new name every time. This is coming from this answer: https://github.com/sequelize/sequelize/issues/9653#issuecomment-660269195

## Checklist 🧢

~[] Documentation~
- [x] Tests
- [x] Ready to be merged

## Affected issues 🎟
None

### DIFF
--- a/packages/sequelize/src/models.ts
+++ b/packages/sequelize/src/models.ts
@@ -26,7 +26,7 @@ export const User = {
     primaryKey: true,
   },
   name: { type: DataTypes.STRING },
-  email: { type: DataTypes.STRING, unique: true },
+  email: { type: DataTypes.STRING, unique: 'email' },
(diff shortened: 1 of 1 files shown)


ndom91:
Great addition, thanks!

ndom91:
For any users that may be coming across this in the future, and still have the issue after updating to a more recent version of the sequelize adapter, which will have this update, remember you may have to manually drop the duplicated indexes (https://github.com/sequelize/sequelize/issues/9653#issuecomment-959790249)

PR #385, opened 2022-02-23
Prevent mongoDB updateUser from adding id twice

### LINKED ISSUE #4030: Prisma Adapter on MongoDB errors out when updating user after login
### Adapter type

@next-auth/prisma-adapter

### Environment

- next: 12.1.0
- next-auth: 4.2.1
- @next-auth/prisma-adapter: 1.0.1
- react: 17.0.2

### Reproduction URL

https://gist.github.com/ferrohd/d769b39194f03844d18e2bcfe85a61e6

### Describe the issue

I'm using PrismaAdapter with MongoDB as underlying database. I don't know if it's because of the prisma.model I'm using, but Prisma doesn't allow to update the "id" field of a document in a collection. When logging-in with an already registered user, the `updateUser` function is called, and tries to update the `emailVerified` field in the User model.

The `data` argument passed to the function has values:
```typescript
{
      id: '62057f64643a91216fcd9f73', // id of the user to update
      emailVerified: 2022-02-21T08:43:23.333Z
}
```
The implementation of updateUser in @next-auth/prisma-adapter ([here](https://github.com/nextauthjs/next-auth/blob/3ec3761634f7953acc52d460e8bb0b2a678d7703/packages/adapter-prisma/src/index.ts#L16)) passes all the data to the update method, including the id
```typescript
updateUser: (data) => p.user.update({ where: { id: data.id }, data })
```
Which causes the following error:
```
Invalid `prisma.user.update()` invocation:

{
  where: {
    id: '62057f64643a91216fcd9f73'
  },
  data: {
    id: '62057f64643a91216fcd9f73',
    ~~
    emailVerified: new Date('2022-02-21T10:58:02.657Z')
  }
}

Unknown arg `id` in data.id for type UserUpdateInput. Available args:

type UserUpdateInput {
  name?: String | NullableStringFieldUpdateOperationsInput | Null
  email?: String | NullableStringFieldUpdateOperationsInput | Null
  emailVerified?: DateTime | NullableDateTimeFieldUpdateOperationsInput | Null
  image?: String | NullableStringFieldUpdateOperationsInput | Null
  token?: Int | IntFieldUpdateOperationsInput
  accounts?: AccountUpdateManyWithoutUserInput
  sessions?: SessionUpdateManyWithoutUserInput
}
```

## Working Fix
```typescript
updateUser: (data) => {
    const { id, ...rest } = data;
    return p.user.update({ where: { id }, data: rest })
}
```

### How to reproduce

The error occur in the `updateUser` function of the adapter. To do so the user must already be registered in the database, otherwise `createUser` is called.
- Register a User (nothing happens, the `createUser` function is called because the user does't exists)
- Logout
- Login with the same email using the EmailProvider

The last step triggers the `updateUser` function of the adapter to update the `emailVerified` field in the User model.

### Expected behavior

Using Prisma with MongoDB should correctly update the user when logging in with the EmailProvider

stophecom:
#### MongoDB Adapter
I noticed that when using the `updateUser` function, the `id` is basically copied and stored as an additional property. This seems unnecessary. This PR fixes this issue for the MongoDB adapter, I haven't checked if the same pattern exists within other adapters. In any case, it is not a big issue, but may save some bytes.

Screen_Shot_2022-02-23_at_22_01_10

## Checklist 🧢

~~- [ ] Documentation~~
~~- [ ] Tests~~
- [x] Ready to be merged

## Affected issues 🎟

Might fix https://github.com/nextauthjs/next-auth/issues/4030

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ All adapter tests should also run against a local instance of the particular dat
 
 ## Publishing
 
-- [Lerna Independent Mode with Semver](https://samhogy.co.uk/2018/08/lerna-independent-mode-with-semver.html)
+- [Lerna Independent Mode with Semver](https://samhogy.co.uk/2018/08/lerna-independent-mode-with-semver)
 
 ## License
 
--- a/packages/mongodb/src/index.ts
+++ b/packages/mongodb/src/index.ts
@@ -111,9 +111,10 @@ export function MongoDBAdapter(
       return from<AdapterUser>(user)
     },
     async updateUser(data) {
+      const { id, ...rest } = data
       const { value: user } = await (
         await db
-      ).U.findOneAndUpdate({ _id: _id(data.id) }, { $set: data })
+      ).U.findOneAndUpdate({ _id: _id(id) }, { $set: rest })
       return from<AdapterUser>(user!)
     },
     async deleteUser(id) {


balazsorban44:
Hi, thank you for the PR, but this repository is not being maintained anymore.

New PRs should be opened against the main repo as stated in the README file.

Here is the adapter source https://github.com/nextauthjs/next-auth/tree/main/packages/adapter-mongodb

Could you reopen it there? :pray:

stophecom:
Sure. True, I somewhat ended up the PR on wrong repo. Here you go https://github.com/nextauthjs/next-auth/pull/4066 :)

